PR #5, opened 2020-02-06
[FLINK-15831] Improve documentation in README.md

patricklucas:
cc @uce

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,44 +1,86 @@
-flink-docker
-============
+Apache Flink Docker Images
+==========================
+
+This repo contains Dockerfiles for building Docker images for Apache Flink, and are used to build
+the "official" [`flink`](https://hub.docker.com/_/flink) images hosted on Docker Hub.
+
+These Dockerfiles are maintained by the Apache Flink community, but the Docker community is
+responsible building and hosting the images on Docker Hub.
 
 [![Build Status](https://travis-ci.org/apache/flink-docker.svg?branch=master)](https://travis-ci.org/apache/flink-docker)
 
-Docker packaging for Apache Flink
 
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -1,44 +1,86 @@
-flink-docker
-============
+Apache Flink Docker Images
+==========================
+
+This repo contains Dockerfiles for building Docker images for Apache Flink, and are used to build
+the "official" [`flink`](https://hub.docker.com/_/flink) images hosted on Docker Hub.
+
+These Dockerfiles are maintained by the Apache Flink community, but the Docker community is
+responsible building and hosting the images on Docker Hub.
```
uce:
```suggestion
responsible for building and hosting the images on Docker Hub.
```

patricklucas:
> LGTM overall. I ran the scripts for a fake 1.10 release as described and the generated output looks good (see [my fork](https://github.com/uce/flink-docker/tree/flink_15831)). Easy to follow process. 👍
> 
> For `generate-stackbrew-library.sh` I noticed two things that we may want to address either as part of this PR or in a follow up:
> 
> ```shell
> declare -A aliases=(
>     [1.9]='latest'
> )
> ```
> 
> Do we want to add a note that this needs to be updated if a new version is added? We can also add logic to figure out the latest version automatically later.
> 
> The call
> 
> ```shell
> versions=( ?.?/ )
> ```
> 
> does not recognize `1.10` and therefore does not generate any output for 1.10.

Good point and nice catch. Maybe for now I'll add separate sections for new Flink minor release and new Flink patch release.

I'd love to just rewrite this in Python or something, using Bash for some of this logic is fairly painful.

PR #10, opened 2020-04-08
[FLINK-17077] Use FLINK_CONF_DIR environment variable to locate flink-conf.yaml

elanv:
To use flink-conf.yaml outside Flink home directory, we should use FLINK_CONF_DIR.
But despite of FLINK_CONF_DIR is set, docker-entrypoint.sh doesn't know the location of which user want. Therefore it would be good to use FLINK_CONF_DIR for the location of flink-conf.yaml, if user provide it.

### DIFF
--- a/docker-entrypoint.sh
+++ b/docker-entrypoint.sh
@@ -20,7 +20,8 @@
 
 # If unspecified, the hostname of the container is taken as the JobManager address
 JOB_MANAGER_RPC_ADDRESS=${JOB_MANAGER_RPC_ADDRESS:-$(hostname -f)}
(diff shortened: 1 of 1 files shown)


tillrohrmann:
Thanks for creating this PR @elanv.

The Flink community is currently discussing how to improve the existing docker images. The effort can be found under FLIP-111. Since this ongoing work is related to your changes, it might make sense to add it once the unification work has completed.

I'm cc'ing also @azagrebin who works on this thread so that he is aware of it.

elanv:
I will try another way as discussed in jira.

PR #12, opened 2020-04-16
 [FLINK-17187] Add utility method for setting config options 

zentol:


docker-entrypoint.sh
```diff
@@ -56,6 +56,21 @@ copy_plugins_if_required() {
   done
 }
 
+set_config_option() {
+  local option=$1
+  local value=$2
+
+  # escape periods for usage in regular expressions
+  local escaped_option=$(echo ${option} | sed -e "s/\./\\\./g")
```
azagrebin, line 64:
```suggestion
  local escaped_option=${option//\./\\\.}
```

docker-entrypoint.sh
```diff
@@ -56,6 +56,21 @@ copy_plugins_if_required() {
...
+
+  # either override an existing entry, or append a new one
+  if grep -E "^${escaped_option}:.*" "${CONF_FILE}}" > /dev/null; then
```
azagrebin, line 67:
```suggestion
  if grep -E "^${escaped_option}:.*" "${CONF_FILE}" > /dev/null; then
```

docker-entrypoint.sh
```diff
@@ -56,6 +56,21 @@ copy_plugins_if_required() {
...
+        sed -i -e "s/${escaped_option}:.*/$option: $value/g" "${CONF_FILE}"
```
azagrebin, line 68: Just some side thoughts:
I suppose the purpose of this is to rewrite the value in the default `flink-conf.yaml` shipped with Flink release.
If we decide to expose an env var to point to custom config location (e.g. a mounted volume)
then we have to reconsider this step or do it before to make sure we rewrite the default `flink-conf.yaml`.

docker-entrypoint.sh
```diff
@@ -64,23 +79,9 @@ elif [ "$1" = "jobmanager" ]; then
...
-    fi
-
-    if grep -E "^query\.server\.port:.*" "${CONF_FILE}" > /dev/null; then
-        sed -i -e "s/query\.server\.port:.*/query.server.port: 6125/g" "${CONF_FILE}"
-    else
-        echo "query.server.port: 6125" >> "${CONF_FILE}"
-    fi
+    set_config_option jobmanager.rpc.address ${JOB_MANAGER_RPC_ADDRESS}
+    set_config_option blob.server.port 6124
+    set_config_option query.server.port 6125
```
azagrebin, line 84: do you want to deduplicate this further? e.g.:
```
copy_plugins_if_required
set_config_option jobmanager.rpc.address ${JOB_MANAGER_RPC_ADDRESS}
set_config_option blob.server.port 6124
set_config_option query.server.port 6125
FLINK_PROPERTIES ...
```
->
```
configure_jm() / common_config() / etc ..
```
or in a later PR?

docker-entrypoint.sh
```diff
@@ -96,29 +97,10 @@ elif [ "$1" = "taskmanager" ]; then
...
-    else
-        echo "query.server.port: 6125" >> "${CONF_FILE}"
-    fi
+    set_config_option jobmanager.rpc.address ${JOB_MANAGER_RPC_ADDRESS}
+    set_config_option taskmanager.numberOfTaskSlots ${TASK_MANAGER_NUMBER_OF_TASK_SLOTS}
+    set_config_option blob.server.port 6124
+    set_config_option query.server.port 6125
 
     if [ -n "${FLINK_PROPERTIES}" ]; then
         echo "${FLINK_PROPERTIES}" >> "${CONF_FILE}"
```
azagrebin, line 106: I guess there is no evil in duplicated config options perse.
Just, maybe, it would be cleaner to iterate over lines in `FLINK_PROPERTIES` and call `set_config_option`
btw, this could be also some `custom_config()` or so

docker-entrypoint.sh
zentol: There is no difference between mounting a custom config and the current `FLINK_PROPERTIES` approach in terms of how we need to handle things.

Take shipped flink-conf.yaml, set various options by default, merge custom config into shipped flink-conf.yaml.

docker-entrypoint.sh
zentol: I suppose we could introduce a shared method for the rpc address and ports.

docker-entrypoint.sh
zentol: That is a separate issue. This PR is purely about de-duplicating code.

docker-entrypoint.sh
zentol: doesn't work; from what I can tell this file isn't run with the bash shell.

zentol:
@azagrebin I've addressed your comments.

PR #15, opened 2020-04-23
[FLINK-17346] De-duplicate process setup

zentol:
De-duplicates the logic for starting JM/TM processes.

### DIFF
--- a/testing/testing_lib.sh
+++ b/testing/testing_lib.sh
@@ -45,51 +45,42 @@ function build_image() {
     docker build -t "$image_name" "$dockerfile_dir"
 }
 
-function run_jobmanager() {
-    local dockerfile
-    dockerfile="$1"
+function internal_run() {
+    local dockerfile="$1"
+    local docker_run_command="$2"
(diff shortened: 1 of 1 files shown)


testing/testing_lib.sh
```diff
@@ -45,51 +45,42 @@ function build_image() {
...
+
+    eval "docker run --rm --detach --network $NETWORK_NAME -e JOB_MANAGER_RPC_ADDRESS=jobmanager ${docker_run_command} $image_name ${args}"
+}
+
+function internal_run_jobmanager() {
+    internal_run "$1" "--name jobmanager --publish 6123:6123 --publish 8081:8081 $2" jobmanager
+}
+
+function run_jobmanager() {
+    internal_run_jobmanager "$1" ""
```
azagrebin, line 67: do we need `""` at the end?

zentol: probably not, but I think it is better to be explicit about it. If another argument is added you're less likely to pass the new argument as the second one.

PR #17, opened 2020-04-24
[FLINK-17367] Add configurable download URL

zentol:
Adds a hook into `add-version.sh` to generate dockerfiles that download the distribution from a configurable URL. Since these likely won't match any GPG signature (as is the case for our nightly snapshots) this check will be skipped.

This required that we pass the URLs directly into the `Dockerfile`, instead of having it generate them from the flink/scala versions.  A new `GPG_SKIP` flag controls whether the signature should be verified.

### DIFF
--- a/Dockerfile-debian.template
+++ b/Dockerfile-debian.template
@@ -46,7 +46,8 @@ RUN set -ex; \
 # Configure Flink version
 ENV FLINK_TGZ_URL=%%BINARY_DOWNLOAD_URL%% \
     FLINK_ASC_URL=%%ASC_DOWNLOAD_URL%% \
-    GPG_KEY=%%GPG_KEY%%
+    GPG_KEY=%%GPG_KEY%% \
+    CHECK_GPG=%%CHECK_GPG%%
 
 # Prepare environment
 ENV FLINK_HOME=/opt/flink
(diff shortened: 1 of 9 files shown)


add-version.sh
```diff
@@ -95,19 +99,46 @@ fi
...
-            -e "s/%%SCALA_VERSION%%/$scala_version/" \
-            -e "s/%%GPG_KEY%%/$gpg_key/" \
-            "Dockerfile-$source_variant.template" > "$dir/Dockerfile"
+        generate "${dir}" "${flink_tgz_url}" "${flink_asc_url}" ${gpg_key} false ${source_variant}
     done
+
+    if [ -n "${binary_download_url}" ]; then
+        dir="$flink_release/custom-${source_variant}"
+
+        generate "${dir}" "${binary_download_url}" "" "" true ${source_variant}
```
azagrebin, line 141: a bit weird that the script has to always generate both `flink_version` and `binary_download_url`
and not either `flink_version` or `binary_download_url`.
if we need `binary_download_url`, do we always want to generate a normal `flink_version` as well?

Dockerfile-debian.template
```diff
@@ -55,27 +56,24 @@ RUN groupadd --system --gid=9999 flink && \
...
-                          hkp://p80.pool.sks-keyservers.net:80 \
-                          keyserver.ubuntu.com \
-                          hkp://keyserver.ubuntu.com:80 \
-                          pgp.mit.edu) ; do \
-      gpg --batch --keyserver "$server" --recv-keys "$GPG_KEY" && break || : ; \
-  done && \
-  gpg --batch --verify flink.tgz.asc flink.tgz; \
-  gpgconf --kill all; \
-  rm -rf "$GNUPGHOME" flink.tgz.asc; \
+  if [ "$SKIP_GPG" = "false" ]; then \
```
azagrebin, line 63: nit: maybe a matter of personal taste to not use negated vars, I would consider a `CHECK_GPG = true` flag.

add-version.sh
zentol: That's a fair point. I did not want to change any existing code path (to ensure it still works), but at the same time not introduce multiple files; so we end up with this mixed bag.

I'll try an approach with separate files.

zentol:
@azagrebin Please take another look.

add-custom.sh
```diff
@@ -0,0 +1,49 @@
...
+if [ -z "${binary_download_url}" ]; then
+    usage
+    exit 1
+fi
+
+mkdir -p "dev"
+
+echo -n >&2 "Generating Dockerfiles..."
+for source_variant in "${SOURCE_VARIANTS[@]}"; do
+  dir="dev/$name-${source_variant}"
```
azagrebin, line 44: nit
```suggestion
  dir="dev/${name}-${source_variant}"
```
also in other places.
afaik, we do not have code style for this but it might be good to follow one style.

PR #19, opened 2020-04-30
[FLINK-17164] Add job-cluster support

zentol:
Adds support for job-clusters.

This PR is also relevant for 1.10, but the commits would be identical.

A job-cluster can be started by passing `standalone-job` to `docker-entrypoint.sh`. User artifacts (i.e., the job jar to run) can be passed via the `USER_ARTIFACTS` environment variable as a list of files/directories, separated with semi-colons (;). This will usually point to some mounted volume.
The script then copies all files into the `usrlib` directory of the distribution.

 The vast majority of this PR is concerned with testing. Since a job-cluster cannot be started without a job to run I added a maven project containing a trivial streaming job that runs infinitely.
The jar is built on CI before the tests are run and provided to the jobmanager as described above.

### DIFF
--- a/docker-entrypoint.sh
+++ b/docker-entrypoint.sh
@@ -18,6 +18,8 @@
 # limitations under the License.
 ###############################################################################
 
+COMMAND_STANDALONE="standalone-job"
+
 # If unspecified, the hostname of the container is taken as the JobManager address
 JOB_MANAGER_RPC_ADDRESS=${JOB_MANAGER_RPC_ADDRESS:-$(hostname -f)}
 CONF_FILE="${FLINK_HOME}/conf/flink-conf.yaml"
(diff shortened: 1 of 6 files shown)


docker-entrypoint.sh
```diff
@@ -56,6 +58,27 @@ copy_plugins_if_required() {
   done
 }
 
+setup_usrlib() {
+  if [ -z "${USER_ARTIFACTS}" ]; then
+    return 0
```
zentol, line 63: one could argue that we should fail here immediately; but I wasn't sure.

testing/run_travis_tests.sh
```diff
@@ -12,6 +12,8 @@ fi
 BRANCH="$TRAVIS_BRANCH"
 
 if [ -n "$IS_PULL_REQUEST" ]; then
+  # fetch the branch the PR opened against; we can only generate a diff against master by default
```
zentol, line 15:
```suggestion
  # fetch the branch the PR was opened against; we can only generate a diff against master by default
```

zentol:
Have to sort out some travis dependencies first (java+maven)

PR #20, opened 2020-04-30
[FLINK-17164] Add job-cluster support 

zentol:
Adds support for job-clusters.

This PR is also relevant for 1.10, but the commits would be identical.

A job-cluster can be started by passing standalone-job to docker-entrypoint.sh. User artifacts (i.e., the job jar to run) can be passed via the USER_ARTIFACTS environment variable as a list of files/directories, separated with semi-colons (;). This will usually point to some mounted volume.
The script then copies all files into the usrlib directory of the distribution.

The vast majority of this PR is concerned with testing. Since a job-cluster cannot be started without a job to run I added a maven project containing a trivial streaming job that runs infinitely.
The jar is built on CI before the tests are run and provided to the jobmanager as described above.

This PR also removes the diff generation from the testing scripts, which printed the names of all modified names. It doesn't work as is since we no longer compare against the master branch, and I do not consider it valuable enough to invest additional time into fixing it.

### DIFF
--- a/docker-entrypoint.sh
+++ b/docker-entrypoint.sh
@@ -18,6 +18,8 @@
 # limitations under the License.
 ###############################################################################
 
+COMMAND_STANDALONE="standalone-job"
+
 # If unspecified, the hostname of the container is taken as the JobManager address
 JOB_MANAGER_RPC_ADDRESS=${JOB_MANAGER_RPC_ADDRESS:-$(hostname -f)}
 CONF_FILE="${FLINK_HOME}/conf/flink-conf.yaml"
@@ -77,11 +79,7 @@ set_common_options() {
     set_config_option query.server.port 6125
 }
 
-if [ "$1" = "help" ]; then
-    echo "Usage: $(basename "$0") (jobmanager|taskmanager|help)"
-    exit 0
-elif [ "$1" = "jobmanager" ]; then
-    shift 1
+prepare_job_manager_start() {
     echo "Starting Job Manager"
     copy_plugins_if_required
 
@@ -91,8 +89,21 @@ elif [ "$1" = "jobmanager" ]; then
         echo "${FLINK_PROPERTIES}" >> "${CONF_FILE}"
     fi
     envsubst < "${CONF_FILE}" > "${CONF_FILE}.tmp" && mv "${CONF_FILE}.tmp" "${CONF_FILE}"
+}
+
+if [ "$1" = "help" ]; then
+    echo "Usage: $(basename "$0") (jobmanager|${COMMAND_STANDALONE}|taskmanager|help)"
+    exit 0
+elif [ "$1" = "jobmanager" ]; then
+    shift 1
+    prepare_job_manager_start
 
     exec $(drop_privs_cmd) "$FLINK_HOME/bin/jobmanager.sh" start-foreground "$@"
+elif [ "$1" = ${COMMAND_STANDALONE} ]; then
+    shift 1
+    prepare_job_manager_start
+
+    exec $(drop_privs_cmd) "$FLINK_HOME/bin/standalone-job.sh" start-foreground "$@"
 elif [ "$1" = "taskmanager" ]; then
     shift 1
     echo "Starting Task Manager"
--- a/testing/docker-test-job/pom.xml
+++ b/testing/docker-test-job/pom.xml
@@ -1,0 +1,75 @@
+<!--
+Licensed to the Apache Software Foundation (ASF) under one
+or more contributor license agreements.  See the NOTICE file
+distributed with this work for additional information
+regarding copyright ownership.  The ASF licenses this file
+to you under the Apache License, Version 2.0 (the
+"License"); you may not use this file except in compliance
+with the License.  You may obtain a copy of the License at
+
+  http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing,
+software distributed under the License is distributed on an
+"AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+KIND, either express or implied.  See the License for the
+specific language governing permissions and limitations
+under the License.
+-->
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+	xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+	<modelVersion>4.0.0</modelVersion>
+
+	<groupId>org.apache.flink</groupId>
+	<artifactId>docker-test-job</artifactId>
+	<version>1.0</version>
+	<packaging>jar</packaging>
+
+	<properties>
+		<project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+		<flink.version>1.9.0</flink.version>
+		<java.version>1.8</java.version>
+		<scala.binary.version>2.11</scala.binary.version>
+		<maven.compiler.source>${java.version}</maven.compiler.source>
+		<maven.compiler.target>${java.version}</maven.compiler.target>
+	</properties>
+
+	<dependencies>
+		<dependency>
+			<groupId>org.apache.flink</groupId>
+			<artifactId>flink-streaming-java_${scala.binary.version}</artifactId>
+			<version>${flink.version}</version>
+			<scope>provided</scope>
+		</dependency>
+	</dependencies>
+
+	<build>
+		<plugins>
+
+			<!-- Java Compiler -->
+			<plugin>
+				<groupId>org.apache.maven.plugins</groupId>
+				<artifactId>maven-compiler-plugin</artifactId>
+				<version>3.1</version>
+				<configuration>
+					<source>${java.version}</source>
+					<target>${java.version}</target>
+				</configuration>
+			</plugin>
+
+			<plugin>
+				<groupId>org.apache.maven.plugins</groupId>
+				<artifactId>maven-shade-plugin</artifactId>
+				<version>3.1.1</version>
+				<executions>
+					<execution>
+						<phase>package</phase>
+						<goals>
+							<goal>shade</goal>
+						</goals>
+					</execution>
+				</executions>
+			</plugin>
+		</plugins>
+	</build>
+</project>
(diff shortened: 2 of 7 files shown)


docker-entrypoint.sh
```diff
@@ -18,6 +18,8 @@
 # limitations under the License.
 ###############################################################################
 
+COMMAND_STANDALONE="standalone-job"
```
azagrebin, line 21: Is `standalone-job` already a stable name for the single job mode because of its user-facing bin script?
I personally find it confusing. What about `single-jobmanager` and add another alias for cluster mode: `cluster-jobmanager`?

docker-entrypoint.sh
```diff
@@ -56,6 +58,27 @@ copy_plugins_if_required() {
...
+setup_usrlib() {
+  if [ -z "${USER_ARTIFACTS}" ]; then
+    return 0
+  fi
+
+  echo "Setting up usrlib"
+  for user_artifact in $(echo "${USER_ARTIFACTS}" | tr ';' ' '); do
+    echo "Adding artifact ${user_artifact} to usrlib directory."
+
+    mkdir -p "${FLINK_HOME}/usrlib"
```
azagrebin, line 70:
```suggestion
  mkdir -p "${FLINK_HOME}/usrlib"
  for user_artifact in $(echo "${USER_ARTIFACTS}" | tr ';' ' '); do
    echo "Adding artifact ${user_artifact} to usrlib directory."

```
moving `mkdir -p "${FLINK_HOME}/usrlib"` out of loop

testing/docker-test-job/pom.xml
```diff
@@ -0,0 +1,75 @@
...
+	<modelVersion>4.0.0</modelVersion>
+
+	<groupId>org.apache.flink</groupId>
+	<artifactId>docker-test-job</artifactId>
+	<version>1.0</version>
+	<packaging>jar</packaging>
+
+	<properties>
+		<project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+		<flink.version>1.9.0</flink.version>
```
azagrebin, line 30: is it `1.9` because it does not matter too much?
maybe one more place speaking for having flink version somewhere in one place
I suppose it could be rewritten while building it in `build_test_job()`

testing/docker-test-job/pom.xml
```diff
@@ -0,0 +1,75 @@
...
+		<java.version>1.8</java.version>
```
azagrebin, line 31: also, there is going to be Java 11

testing/testing_lib.sh
```diff
@@ -178,17 +182,30 @@ function internal_smoke_test_images() {
...
+        jobmanager_container_id="$(internal_run_jobmanager \
+            "$dockerfile" \
+            "${docker_run_command_args} --mount type=bind,src=$(pwd)/testing/docker-test-job/target,target=/jars -e USER_ARTIFACTS=/jars" \
+            "standalone-job --job-classname org.apache.flink.StreamingJob")"
+        taskmanager_container_id="$(internal_run_taskmanager \
+            "$dockerfile" \
+            "${docker_run_command_args} --mount type=bind,src=$(pwd)/testing/docker-test-job/target,target=/jars")"
+        wait_for_jobmanager "$dockerfile"
+        test_image "$dockerfile"
+        docker kill "$jobmanager_container_id" "$taskmanager_container_id" > /dev/null
```
azagrebin, line 208: could it be deduplicated for cluster/job mode?
```sh
docker_run_command_args="${docker_run_command_args} --mount type=bind,src=$(pwd)/testing/docker-test-job/target,target=/jars -e USER_ARTIFACTS=/jars"
local jm_args="standalone-job --job-classname org.apache.flink.StreamingJob"
test_jm_tm ${dockerfile} ${docker_run_command_args} ${jm_args}
```

testing/run_travis_tests.sh
```diff
@@ -11,14 +11,6 @@ fi
 
 BRANCH="$TRAVIS_BRANCH"
 
-if [ -n "$IS_PULL_REQUEST" ]; then
-  changed_files="$(git diff --name-only $BRANCH...HEAD)"
```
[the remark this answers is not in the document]
zentol: You can't do a diff against an arbitrary branch right after a clone; we'd first need to fetch other branches which I tried but the damn thing still wasn't working.

So I deleted it because this is not something I want to spend time on.
Note that this also applies to the other dev branches.

docker-entrypoint.sh
zentol: Is the best term we got at the moment, and I'd prefer to not introduce another one.

testing/docker-test-job/pom.xml
zentol: it doesn't matter, FWIW it could also be 1.4.0. There is no requirement for this version to be in sync with the Flink version.

testing/docker-test-job/pom.xml
zentol: Yep, we'll handle that when the time comes.

testing/run_travis_tests.sh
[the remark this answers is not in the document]
azagrebin: alright, indeed, the value seems to be not so big comparing to dev&maintenance.

docker-entrypoint.sh
azagrebin: btw, we already called it differently [here](https://github.com/apache/flink/blob/master/flink-container/docker/docker-entrypoint.sh#L24)

tillrohrmann: What about `application` or so? https://ci.apache.org/projects/flink/flink-docs-stable/concepts/glossary.html#flink-application-cluster

docker-entrypoint.sh
```diff
@@ -56,6 +58,27 @@ copy_plugins_if_required() {
   done
 }
 
+setup_usrlib() {
+  if [ -z "${USER_ARTIFACTS}" ]; then
```
azagrebin, line 62: maybe `JOB_ARTIFACTS` or `USER_JOB_ARTIFACTS`?

docker-entrypoint.sh
```diff
@@ -56,6 +58,27 @@ copy_plugins_if_required() {
...
+    return 0
+  fi
+
+  echo "Setting up usrlib"
+  mkdir -p "${FLINK_HOME}/usrlib"
+  for user_artifact in $(echo "${USER_ARTIFACTS}" | tr ';' ' '); do
```
azagrebin, line 68: any specific reason to use `;` separation? not `,` or `:`?

zentol: pretty much copied from the plugins equivalent above.

docker-entrypoint.sh
zentol: @tillrohrmann I wouldn't use `application`; while the glossary contained the concept of `Application Cluster` for years I don't think I have ever seen it mentioned anywhere (especially not anywhere in flink-dist).
The current(master) glossary also says that a Flink _application_ can be both submitted to a session or application cluster, so it isn't exactly clear what `application` would do. Maybe this will change when FLIP-85 (Flink Application Mode) is completed, _if_ it fully subsumes per-job clusters, but until then I would stay away from this term.
Considering the potential future features outlined in the FLIP,
> Being able to launch one image in Kubernetes and the launched containers will figure out who is the JM and who are the TMs.

the semantics of `application` could change again in the future, and it may be best to reserve this term for that time.

@azagrebin I did not know that. I'd still stick with `standalone-job` though to stay close to flink-dist. I very much want to prevent us from introducing separate terminology into the docker images.

docker-entrypoint.sh
zentol: We could generalize this more and call it `USER_LIB_ARTIFACTS` and use it also in session mode.

docker-entrypoint.sh
tillrohrmann: I see your point @zentol that you want to stick to the shell script names.

I guess my main concern is that I associate standalone with a bare metal deployment and not with Docker. Strictly speaking there is not much of a difference though.

Maybe we should rename the `standalone-job.sh` into something more aligned with the glossary. But I agree that this should/could be a separate effort.

azagrebin: My main concern was that it will now become part of documented `docker run flink API`. We can still rename it in future of course and be backwards-compatible with the old name so I am ok with `standalone-job` either.

docker-entrypoint.sh
azagrebin: Not sure about mixing two things. do you mean handle `USER_LIB_ARTIFACTS` differently, depending on job mode: copying to `usrlib` (job) or `lib`(session)? then it is also confusing for tm mode if I understood correctly.

zentol: is there a use-case for putting something into /lib for job clusters?

My thought was that users shouldn't have to worry whether something is in lib or usrlib, but this only works if usrlib is _not_ loaded by a separate user-code classloader.

zentol: Maybe we should a it more long-term with the naming.

Ultimately, what we're doing here is mounting the `usrlib` directory. We want to expose the same for the `conf` directory, so how would we name that one? The should follow the same naming pattern.

zentol:
@azagrebin As discussed in our offline chat, I have removed the environment variable, and we now assume that users mount the usrlib directory to /opt/flink/usrlib.

docker-entrypoint.sh
```diff
@@ -91,8 +89,21 @@ elif [ "$1" = "jobmanager" ]; then
...
+elif [ "$1" = "jobmanager" ]; then
+    shift 1
+    prepare_job_manager_start
 
     exec $(drop_privs_cmd) "$FLINK_HOME/bin/jobmanager.sh" start-foreground "$@"
+elif [ "$1" = ${COMMAND_STANDALONE} ]; then
+    shift 1
+    prepare_job_manager_start
+
+    exec $(drop_privs_cmd) "$FLINK_HOME/bin/standalone-job.sh" start-foreground "$@"
```
azagrebin, line 106: Maybe it would be nice to check that `usrlib` exists and not empty, giving a hint otherwise that the user probably wants to prepare it accordingly. If it does not hold, it will fail later and maybe not so obvious.

zentol: hmm yes good point, but we could only check this for the jobmanager start; taskmanagers may not require `usrlib` after all if they run as part of a session cluster.
This should be something that the standalone-job.sh should be checking though I think, not the docker scripts.

azagrebin: true, I just did not find it there but would be a nice follow-up

PR #23, opened 2020-05-18
[FLINK-14241][test]Add arm64 support for docker e2e test -- work in progress

wangxiyuan:
The docker image `openjdk:8-jre` only works for amd64.

When running  test on arm64, use `arm64v8/openjdk:8-jre` instead.

### DIFF
--- a/generator.sh
+++ b/generator.sh
@@ -15,6 +15,7 @@ function generateDockerfile {
     java_version=$6
     source_variant=$7
 
+    # TODO: Remove this aarch64 condition once all required docker images support arm64 originally.
     if [[ `uname -i` == 'aarch64' ]]; then
         from_docker_image="arm64v8\/openjdk:${java_version}-jre"
     else
--- a/generator.sh
+++ b/generator.sh
@@ -16,7 +16,7 @@ function generateDockerfile {
     source_variant=$7
 
     if [[ `uname -i` == 'aarch64' ]]; then
-        from_docker_image="arm64v8/openjdk:${java_version}-jre"
+        from_docker_image="arm64v8\/openjdk:${java_version}-jre"
     else
         from_docker_image="openjdk:${java_version}-jre"
     fi
--- a/generator.sh
+++ b/generator.sh
@@ -15,7 +15,11 @@ function generateDockerfile {
     java_version=$6
     source_variant=$7
 
-    from_docker_image="openjdk:${java_version}-jre"
+    if [[ `uname -i` == 'aarch64' ]]; then
+        from_docker_image="arm64v8/openjdk:${java_version}-jre"
+    else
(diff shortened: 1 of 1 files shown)


wangxiyuan:
@zentol can you take a look?  Thanks very much

wangxiyuan:
@rmetzger could you please take a look? This block arm64 test for a while. Thanks.

rmetzger:
I'm not sure if this PR works as-is. The updated generator doesn't seem to set the architecture flag correctly: https://github.com/apache/flink-docker/pull/23/files#diff-0c5f825c10334c2eff0a90687f513c69R93

also, the architecture seems to depend on the architecture of the machine executing the `generate.sh` script. But this script only generates the Dockerfile and metadata file, not the actual docker image.

wangxiyuan:
> I'm not sure if this PR works as-is. The updated generator doesn't seem to set the architecture flag correctly: https://github.com/apache/flink-docker/pull/23/files#diff-0c5f825c10334c2eff0a90687f513c69R93
> 
> also, the architecture seems to depend on the architecture of the machine executing the `generate.sh` script. But this script only generates the Dockerfile and metadata file, not the actual docker image.

Acutally many docker images support multi arch currently.  For example `openjdk:11-jre`


https://github.com/wangxiyuan/flink-docker/blob/dev-master/generator.sh#L18-L22

rmetzger:
I'm sorry that I have to say it again, but I don't think the current approach is correct.

The `generate.sh` script generates a Dockerfile and metadata file, then this file is used to create a stackbrew file, which is committed here: https://github.com/docker-library/official-images/blob/master/library/flink.
Then DockerHub does some magic to turn this into the official images.

If we want to add support for arm64 to the official Flink docker images, we probably need to generate an additional Dockerfile and metadata file with an updated FROM, and a correct `release.metadata` file.

I believe that your change works on ARM64 with the Flink e2e tests, but this repository is mostly used for building the official Flink docker images, and we need to make sure we are not breaking them with this change.

wangxiyuan:
Oh, I see. I didn't realize that this is used for offcial image build as well.

Yeah, If so, this PR is not correct. I'll find some way  to only apply for e2e test. Thanks!!

rmetzger:
Great! Thanks!

nferrario:
RocksDB is not currently being built for ARM. We had to manually compile it and override it in the classpath. Kinda hacky but it's the only solution we have until Flink supports it officially.

rmetzger:
The latest frocksdb version Flink is using (https://search.maven.org/artifact/com.ververica/frocksdbjni/6.20.3-ververica-1.0/jar) supports Linux aarch64, so you should be able to run Flink on Docker on an M1 mac book (because Docker for Mac is effectively running in a (arm) Linux VM)

PR #31, opened 2020-07-27
[FLINK-16260] Change generate-stackbrew-library.sh to support new release.metadata file

rmetzger:
The final stackbrew file is identical (except for the order of the entries) with the old file.

Once this and #32 are merged, I'll open PRs to update the individual branches and master.

### DIFF
--- a/generate-stackbrew-library.sh
+++ b/generate-stackbrew-library.sh
@@ -11,8 +11,6 @@ set -eu
 self="$(basename "$BASH_SOURCE")"
 cd "$(dirname "$(readlink -f "$BASH_SOURCE")")"
 
-# remove "latest" and any "scala_" tag, unless it is the latest version
-PRUNE_FROM_NON_LATEST_VERSION="^(latest$|scala\_/*)"
 
 # get the most recent commit which modified any of "$@"
 fileCommit() {
@@ -47,16 +45,9 @@ pruneTags() {
         # tagsString contains latest version. keep "latest" tag
         echo $inTagsString
     else
-        # split list of tags, remove anything containing "latest"
-        IFS=', ' read -r -a inTags <<< "$inTagsString"
-        local outString=""
-        for inTag in "${inTags[@]}"; do
-            if [[ $inTag =~ $PRUNE_FROM_NON_LATEST_VERSION ]]; then
-                continue;
-            fi
-            outString="$outString, $inTag"
-        done
-        echo ${outString:2}
+        # remove "latest" and any "scala_" tag, unless it is the latest version
+        # the "scala_" tag has a similar semantic as the "latest" tag in docker registries. 
+        echo $inTagsString | sed -E 's|,(scala\|latest)[-_[:alnum:]]*||g'
     fi
 }
 
@@ -64,7 +55,7 @@ extractValue() {
     local key="$1"
     local file="$2"
     local line=$(cat $file | grep "$key:")
-    echo $line | cut -d ':' -f 2 | xargs # remove key from line, remove whitespace
+    echo $line | cut -d ':' -f 2 | tr -d ' ' # remove key from line, remove whitespace
 }
 
 # get latest flink version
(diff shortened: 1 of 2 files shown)


generate-stackbrew-library.sh
```diff
@@ -41,21 +36,39 @@ dirCommit() {
...
+    local inTagsString=$1
+    local latestVersion=$2
+    if [[ $inTagsString =~ $latestVersion ]]; then
+        # tagsString contains latest version. keep "latest" tag
+        echo $inTagsString
+    else
+        # split list of tags, remove anything containing "latest"
+        IFS=', ' read -r -a inTags <<< "$inTagsString"
+        local outString=""
+        for inTag in "${inTags[@]}"; do
```
zentol, line 53: couldn't this branch be reduced to
`echo $inTagsString | sed 's/, latest[-_[:alnum:]]*//g' | sed 's/, scala[-_[:alnum:]]*//g'`?

generate-stackbrew-library.sh
```diff
@@ -8,16 +8,11 @@
...
-    [1.11]='latest'
-)
-
 self="$(basename "$BASH_SOURCE")"
 cd "$(dirname "$(readlink -f "$BASH_SOURCE")")"
 
-# Identify directories matching '?.?' (e.g. '1.7') and remove trailing slashes
-versions=( ?.?/ ?.??/ )
-versions=( "${versions[@]%/}" )
+# remove "latest" and any "scala_" tag, unless it is the latest version
```
zentol, line 14: mention that there are `scala_2.11` tags that have semantics similar to `latest`

generate-stackbrew-library.sh
rmetzger: Yeah, you are right. Regex can do the job here as well. After a few hours of bash scripting, your brain is sometimes trying to make a problem seem extra complicated :)

generate-stackbrew-library.sh
```diff
@@ -41,21 +36,39 @@ dirCommit() {
...
-                }
-            ' '{}' + \
-            | sort -u \
-            | xargs bashbrew cat --format '[{{ .RepoName }}:{{ .TagName }}]="{{ join " " .TagEntry.Architectures }}"'
-    ) )"
+extractValue() {
+    local key="$1"
+    local file="$2"
+    local line=$(cat $file | grep "$key:")
+    echo $line | cut -d ':' -f 2 | xargs # remove key from line, remove whitespace
```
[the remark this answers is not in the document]
rmetzger: I took xargs because it does not remove "internal" whitespaces (in the value). However, xargs replaces double whitespaces with a single whitespace ... so my whole intention of not modifying the value itself is gone anyways. I'll use `tr`.

rmetzger:
Thank you for your review. I addressed your comments.

generate-stackbrew-library.sh
```diff
@@ -41,21 +34,32 @@ dirCommit() {
     )
 }
 
-getArches() {
-    local repo="$1"; shift
-    local officialImagesUrl='https://github.com/docker-library/official-images/raw/master/library/'
+# Inputs:
+#  - tags: comma-seprated list of image tags
```
zentol, line 38: variable name does not match (I'd use `tags`)

generate-stackbrew-library.sh
```diff
@@ -41,21 +34,32 @@ dirCommit() {
...
-                }
-            ' '{}' + \
-            | sort -u \
-            | xargs bashbrew cat --format '[{{ .RepoName }}:{{ .TagName }}]="{{ join " " .TagEntry.Architectures }}"'
-    ) )"
+extractValue() {
+    local key="$1"
+    local file="$2"
+    local line=$(cat $file | grep "$key:")
+    echo $line | cut -d ':' -f 2 | tr -d ' ' # remove key from line, remove whitespace
```
zentol, line 58: I looked at various entries in https://github.com/docker-library/official-images and they all separated tags with a comma and space `, ` 😢 

We could use `sed` again: `echo $line | sed 's/[-_[:alnum:]]*: //g'`

rmetzger:
Thank you for your thorough review! 

Merging ...

PR #32, opened 2020-07-27
[FLINK-16260] Add support for Java11 in dev-master / introduce new release.metadata file

rmetzger:
The Java11 docker image works.


generator.sh
```diff
@@ -0,0 +1,80 @@
+#!/bin/bash -e
+
+# Defaults, can vary between versions
+export SOURCE_VARIANTS=(java11-debian debian )
+
+export LATEST_SCALA="2.12"
+
+function generate() {
```
zentol, line 8: I'd split this into 2 functions; one for generating the image, one for generating the meta data.
Right now its a function doing 2 things with (pretty much) 2 non-overlapping argument sets.
Then we don't have to deal with it for `add-custom.sh`.

generator.sh
```diff
@@ -0,0 +1,80 @@
...
+        -e "s/%%FROM_IMAGE%%/${from_docker_image}/" \
+        "$source_file.template" > "$dir/Dockerfile"
+
+    ########################################
+    ### generate "release.metadata" file
+    ########################################
+    
+    # docker image tags:
+    java11_suffix=""
+    if [[ $source_variant =~ "java11" ]] ; then
```
zentol, line 53: its a bit weird to encode the java version into the source_variant when the scala version has its own argument.
I'd rather have the java version as an explicit argument, with a dedicated set of `-java8` tags, + it being considered the default java version for the tag set without java versions.

generator.sh
```diff
@@ -0,0 +1,80 @@
...
+    from_docker_image="openjdk:8-jre"
+    if [[ $source_variant =~ "java11" ]] ; then
+        from_docker_image="openjdk:11-jre"
+    fi
+
+    ########################################
+    ### generate "Dockerfile" file
+    ########################################
+
+    # overwrite variable based on $source_variant to support non-debian releases
```
zentol, line 29: this seems to do the opposite; you're hard-coding debian.
It seems overly complicated to add `java11-debian` into `SOURCE_VARIANTS` only to overwrite it later with `debian` anyway.
A separate java version argument would be a simpler solution.

generator.sh
```diff
@@ -0,0 +1,80 @@
+#!/bin/bash -e
+
+# Defaults, can vary between versions
```
zentol, line 3: can be removed now; everything on the dev branches can vary between versions

generator.sh
```diff
@@ -0,0 +1,80 @@
...
+        # "1.2.0-java11"
+        # "1.2-java11"
+        # "latest-java11"
+        tags="$tags, ${flink_version}${java11_suffix}, ${flink_release}${java11_suffix}, latest${java11_suffix}"
+    fi
+
+    echo "Tags: $tags" >> $dir/release.metadata
+
+    # We currently only support amd64 with Flink.
+    echo "Architectures: amd64" >> $dir/release.metadata
```
zentol, line 78: I'm not sure if we can get by with this; if we want ARM support (FLINK-14241) then we'll something more sophisticated. Not saying it has to be the docker magic we had before (which I'd kinda prefer not to use), but we should have a plan on how to support other architectures.

rmetzger: As far as I understood (from here https://github.com/docker-library/official-images#multiple-architectures), we need to make sure that the image we are referring in `FROM` supports all architectures we are specifying here. 
I guess that we don't need to change anything else for additional ARM support.

rmetzger:
Thanks a lot for your valuable feedback! I addressed all comments.

generator.sh
zentol: For ARM support (as shown in #23) we are using a different `FROM` image, which implies that we need different `Architecture` values per image.

add-version.sh
```diff
@@ -98,18 +98,24 @@ fi
...
-        # Not all mirrors have the .asc files
-        flink_asc_url=https://www.apache.org/dist/${flink_url_file_path}.asc
+            flink_tgz_url="https://www.apache.org/dyn/closer.cgi?action=download&filename=${flink_url_file_path}"
+            # Not all mirrors have the .asc files
+            flink_asc_url=https://www.apache.org/dist/${flink_url_file_path}.asc
 
-        generate "${dir}" "${flink_tgz_url}" "${flink_asc_url}" ${gpg_key} true ${source_variant}
+            mkdir "$dir"
+            generateDockerfile "${dir}" "${flink_tgz_url}" "${flink_asc_url}" ${gpg_key} true ${java_version} ${source_variant}
+            generateReleaseMetadata ${dir} ${flink_release} ${flink_version} ${scala_version} ${java_version} ${source_variant}
```
zentol, line 117:
```suggestion
            generateReleaseMetadata "${dir}" ${flink_release} ${flink_version} ${scala_version} ${java_version} ${source_variant}
```

generator.sh
```diff
@@ -0,0 +1,80 @@
...
+    asc_download_url=$3
+    gpg_key=$4
+    check_gpg=$5
+    java_version=$6
+    source_variant=$7
+
+    from_docker_image="openjdk:${java_version}-jre"
+
+    source_file="Dockerfile-debian"
+
```
zentol, line 22:
```suggestion
```

generator.sh
```diff
@@ -0,0 +1,80 @@
...
+    escaped_binary_download_url=$(echo "$binary_download_url" | sed 's/&/\\\&/')
+
+    # generate Dockerfile
+    sed \
+        -e "s,%%BINARY_DOWNLOAD_URL%%,${escaped_binary_download_url}," \
+        -e "s,%%ASC_DOWNLOAD_URL%%,$asc_download_url," \
+        -e "s/%%GPG_KEY%%/$gpg_key/" \
+        -e "s/%%CHECK_GPG%%/${check_gpg}/" \
+        -e "s/%%FROM_IMAGE%%/${from_docker_image}/" \
+        "$source_file.template" > "$dir/Dockerfile"
```
zentol, line 35:
```suggestion
        "Dockerfile-$source_variant.template" > "$dir/Dockerfile"
```

generator.sh
rmetzger: Mh, okay. But still, I believe we don't need to write any code for this now. It seems that the new structure is able to support multiarch builds fairly easy.

generator.sh
```diff
@@ -0,0 +1,80 @@
+#!/bin/bash -e
+
+export SOURCE_VARIANTS=(debian )
+
+export LATEST_SCALA="2.12"
+export LATEST_JAVA="8"
```
zentol, line 6: rename to `DEFAULT_JAVA`? (same for scala)

generator.sh
```diff
@@ -0,0 +1,80 @@
...
+    sed \
+        -e "s,%%BINARY_DOWNLOAD_URL%%,${escaped_binary_download_url}," \
+        -e "s,%%ASC_DOWNLOAD_URL%%,$asc_download_url," \
+        -e "s/%%GPG_KEY%%/$gpg_key/" \
+        -e "s/%%CHECK_GPG%%/${check_gpg}/" \
+        -e "s/%%FROM_IMAGE%%/${from_docker_image}/" \
+        "$source_file.template" > "$dir/Dockerfile"
+}
+
+# generate "release.metadata" file
```
zentol, line 38:
```suggestion
```

rmetzger:
All items addressed.

rmetzger:
There's one problem with this state:
```
Tags: 1.11.0-scala_2.11-java8, 1.11-scala_2.11-java8, scala_2.11-java8
Architectures: amd64
GitCommit: 4d0495116bfd4f18fe5932197bb5b6f1da90965c
Directory: ./1.11/scala_2.11-java8-debian
```
is missing the variants w/o `-java8`.

rmetzger:
The problem is resolved.

add-custom.sh
```diff
@@ -43,7 +43,7 @@ echo -n >&2 "Generating Dockerfiles..."
 for source_variant in "${SOURCE_VARIANTS[@]}"; do
   dir="dev/${name}-${source_variant}"
   rm -rf "${dir}"
-
-  generate "${dir}" "${binary_download_url}" "" "" false ${source_variant}
+  mkdir "$dir"
+  generateDockerfile "${dir}" "${binary_download_url}" "" "" false 8 ${source_variant}
```
iemejia, line 47: Just looking quickly, this 8 should not be like ${java_version} ?

rmetzger: You are right, it would be nice to introduce a parameter for setting the Java version. I was too lazy adding it 🙈.
I'll add with the next round of comments from Chesnay.

zentol: A separate parameter makes sense to me, but then please also adjust the tests to cover both java 8 and 11.

testing/run_travis_tests.sh
```diff
@@ -11,14 +11,25 @@ fi
 
 BRANCH="$TRAVIS_BRANCH"
 
-./add-custom.sh -u "https://s3.amazonaws.com/flink-nightly/flink-1.11-SNAPSHOT-bin-hadoop2.tgz"
-
-if [ -z "$IS_PULL_REQUEST" ] && [ "$BRANCH" = "dev-master" ]; then
+function run_tests {
+	if [ -z "$IS_PULL_REQUEST" ] && [ "$BRANCH" = "dev-master" ]; then
```
zentol, line 15: this file is indented with spaces

testing/run_travis_tests.sh
```diff
@@ -11,14 +11,25 @@ fi
...
 else
   # For pull requests and branches, test one image
   smoke_test_one_image
 fi
+}
+
+./add-custom.sh -u "https://s3.amazonaws.com/flink-nightly/flink-1.11-SNAPSHOT-bin-hadoop2.tgz"
+run_tests
+
+rm -r dev
```
zentol, line 27: Alternatively, you can give the custom builds a name (using `-n`), build both, and then just run `smoke_test_all_images`.
This whole setup of running 1 image per PR was imo mostly relevant for the previous setup where tests ran on the master branch with many images sitting in the branch.

rmetzger:
Thanks! Merging ...

generator.sh
wangxiyuan: `openjdk:11-jre` supports arm64 by default. 

But `openjdk:8-jre` doesn't.  https://hub.docker.com/layers/openjdk/library/openjdk/8-jre/images/sha256-e14239f5b4190a136e576aa4f39b645ef683732f628595e1e16a0c0595752fb1?context=explore

So it still needs to get the arm64 image from `arm64v8` repo currently.
https://github.com/apache/flink-docker/pull/23

PR #33, opened 2020-07-30
Add support for Java 11 and the new "release.metadata" file to dev-1.11

rmetzger:


iemejia:
We can add this retroactively to 1.10 too if there is interest too.

rmetzger:
Yeah, but we shouldn't make people too comfortable using old Flink versions :) But I agree, if there's strong demand, we can consider it.

I'm going to merge this change now! Thanks for the reviews!

PR #36, opened 2020-07-30
[hotfix] Replace forbidden latest-javaX with javaX

rmetzger:
The PR adding the release was rejected by their CI because of this: https://github.com/docker-library/official-images/pull/8459

### DIFF
--- a/generator.sh
+++ b/generator.sh
@@ -71,7 +71,7 @@ function generateReleaseMetadata {
         # we are generating the image for the latest scala version, add:
         # "1.2.0-java11"
         # "1.2-java11"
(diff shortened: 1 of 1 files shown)


rmetzger:
Once this is approved, I'll apply it to "dev-1.11" as well (w/o PR).

generator.sh
```diff
@@ -72,7 +72,7 @@ function generateReleaseMetadata {
         # "1.2.0-java11"
         # "1.2-java11"
         # "latest-java11"
```
zentol, line 74:
```suggestion
        # "java11"
```

PR #37, opened 2020-08-10
[FLINK-15793] Support native K8s mode

wangyang0918:
This PR will integrate flink`kubernetes-entry.sh` into the unified flink-docker `docker-entrypoint.sh`. After then users can enable build-in plugins much more easily. 

Using the following command to enable the plugin.
```
flink run-application -t kubernetes-application -Dkubernetes.cluster-id=k8s-ha-app1 ...
-Dcontainerized.master.env.ENABLE_BUILT_IN_PLUGINS=flink-oss-fs-hadoop-1.11.1.jar \
-Dcontainerized.taskmanager.env.ENABLE_BUILT_IN_PLUGINS=flink-oss-fs-hadoop-1.11.1.jar \
local:///opt/flink/examples/streaming/StateMachineExample.jar
```

# How to verify this PR?
Update Flink code
* Use `docker-entrypoint.sh` as entrypoint of JobManager and TaskManager pod
* Add `native-k8s` as an argument

Then run a Flink native K8s session or application, it should deploy successfully just as before.


wangyang0918:
cc @rmetzger Could help with the review?

docker-entrypoint.sh
```diff
@@ -120,6 +120,16 @@ elif [ "$1" = "taskmanager" ]; then
...
+elif [ "$1" = "native-k8s" ]; then
+    shift 1
+    copy_plugins_if_required
+
+    export _FLINK_HOME_DETERMINED=true
+    . $FLINK_HOME/bin/config.sh
+    export FLINK_CLASSPATH="`constructFlinkClassPath`:$INTERNAL_HADOOP_CLASSPATHS"
+    # Start commands for jobmanager and taskmanager are generated by Flink internally.
+    echo "Start command: $*"
+    exec bash -c "$@"
```
rmetzger, line 132: Any reason why you are not using `drop_privs_cmd` like the other commands?
Many enterprise k8s environments require non-root container support, and it seems easy to support for us.

docker-entrypoint.sh
```diff
@@ -120,6 +120,16 @@ elif [ "$1" = "taskmanager" ]; then
     envsubst < "${CONF_FILE}" > "${CONF_FILE}.tmp" && mv "${CONF_FILE}.tmp" "${CONF_FILE}"
 
     exec $(drop_privs_cmd) "$FLINK_HOME/bin/taskmanager.sh" start-foreground "$@"
+elif [ "$1" = "native-k8s" ]; then
```
rmetzger, line 123: Why are did you not add this command to the output of the "help" command?

docker-entrypoint.sh
wangyang0918: I am just keeping the same behavior with `kubernetes-entry.sh`. I think you are right. We need to do `drop_privs_cmd` just like other commands.

docker-entrypoint.sh
wangyang0918: Will add `native-k8s` to help message.

wangyang0918:
@rmetzger I have added a `fixup` commit to address the comments.

rmetzger:
Thanks. The change looks good!

Merging ...

PR #38, opened 2020-08-21
Add GPG key for 1.10.2 release and update test script

zhuzhurk:


zhuzhurk:
@rmetzger would you help to take a look at this pull request?

zhuzhurk:
Travis CI failed because the used mirror did not catch up with the newly released 1.10.2.
Will update the commit later to trigger another travis CI.

PR #40, opened 2020-09-16
Add GPG key for 1.11.2 release and update test script

zhuzhurk:


zhuzhurk:
The docker image building cannot succeed because `wget` fails to fetch from the binary url[1]. The error is "unknown action [download]".
However, the url works in web browser.
@rmetzger @zentol do you have any idea why this could happen? Or is there any replacement for "https://www.apache.org/dyn/closer.cgi" ?

[1] wget -nv -O flink.tgz https://www.apache.org/dyn/closer.cgi?action=download&filename=flink/flink-1.11.2/flink-1.11.2-bin-scala_2.11.tgz

rmetzger:
I think the problem are the = and & characters in the download URL.

If you quote the URL, it works: wget -nv -O flink.tgz "https://www.apache.org/dyn/closer.cgi?action=download&filename=flink/flink-1.11.2/flink-1.11.2-bin-scala_2.11.tgz"

zhuzhurk:
Thanks for the information @rmetzger !
Yes you are right `wget` can work with quoted url .
However, I checked the generated `Dockerfile` and the command is `wget -nv -O flink.tgz "$FLINK_TGZ_URL"; `. It is in a good format so seems it's just the logged command which is not correct. And the root cause of travis failure might not be the one I guessed.
I will try to re-trigger travis to see if the problem still exists.

zhuzhurk:
Travis gives green now. Seems the problem is gone.

rmetzger (approved):
Thanks a lot.

+1 to merge.

PR #42, opened 2020-11-11
[FLINK-19125] Adopt Jemalloc as default memory allocator to avoid memory leak

Myasuka:
This PR is still a draft PR for pre-review until 1.12 docker image folder is created and we plan to apply changes on 1.12 docker images.

### DIFF
--- a/1.11/scala_2.11-java11-debian/docker-entrypoint.sh
+++ b/1.11/scala_2.11-java11-debian/docker-entrypoint.sh
@@ -91,21 +91,47 @@ prepare_job_manager_start() {
     envsubst < "${CONF_FILE}" > "${CONF_FILE}.tmp" && mv "${CONF_FILE}.tmp" "${CONF_FILE}"
 }
 
+disable_jemalloc_env() {
+  if [ "$1" = "disablejemalloc" ]; then
+      echo "Disable Jemalloc as memory allocator" 
+      return 0
+  else
+      echo "Enable Jemalloc via appending env variable LD_PRELOAD with /usr/lib/x86_64-linux-gnu/libjemalloc.so"
+      export LD_PRELOAD=$LD_PRELOAD:/usr/lib/x86_64-linux-gnu/libjemalloc.so
+      return 1
+  fi
+}
+
 if [ "$1" = "help" ]; then
     echo "Usage: $(basename "$0") (jobmanager|${COMMAND_STANDALONE}|taskmanager|help)"
     exit 0
 elif [ "$1" = "jobmanager" ]; then
     shift 1
+    disable_jemalloc_env $@
+    disabled_jemalloc="$?"
+    if [ ${disabled_jemalloc} = 0 ]; then 
+      shift 1
+    fi
     prepare_job_manager_start
 
     exec $(drop_privs_cmd) "$FLINK_HOME/bin/jobmanager.sh" start-foreground "$@"
 elif [ "$1" = ${COMMAND_STANDALONE} ]; then
     shift 1
+    disable_jemalloc_env $@
+    disabled_jemalloc="$?"
+    if [ ${disabled_jemalloc} = 0 ]; then 
+      shift 1
+    fi
     prepare_job_manager_start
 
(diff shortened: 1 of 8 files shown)


1.11/scala_2.11-java11-debian/docker-entrypoint.sh
```diff
@@ -91,21 +91,47 @@ prepare_job_manager_start() {
     envsubst < "${CONF_FILE}" > "${CONF_FILE}.tmp" && mv "${CONF_FILE}.tmp" "${CONF_FILE}"
 }
 
+disable_jemalloc_env() {
+  if [ "$1" = "disablejemalloc" ]; then
```
tillrohrmann, line 95: I think we should add somewhere in the usage description that `disablejemalloc` is an option and describe what it does.

1.11/scala_2.11-java11-debian/docker-entrypoint.sh
```diff
@@ -91,21 +91,47 @@ prepare_job_manager_start() {
...
+      return 0
+  else
+      echo "Enable Jemalloc via appending env variable LD_PRELOAD with /usr/lib/x86_64-linux-gnu/libjemalloc.so"
+      export LD_PRELOAD=$LD_PRELOAD:/usr/lib/x86_64-linux-gnu/libjemalloc.so
+      return 1
+  fi
+}
+
 if [ "$1" = "help" ]; then
     echo "Usage: $(basename "$0") (jobmanager|${COMMAND_STANDALONE}|taskmanager|help)"
```
tillrohrmann, line 106: The usage description is here.

1.11/scala_2.11-java11-debian/docker-entrypoint.sh
```diff
@@ -91,21 +91,47 @@ prepare_job_manager_start() {
     envsubst < "${CONF_FILE}" > "${CONF_FILE}.tmp" && mv "${CONF_FILE}.tmp" "${CONF_FILE}"
 }
 
+disable_jemalloc_env() {
```
tillrohrmann, line 94: This method seems to be duplicated across all `docker-entrypoint.sh` scripts. I guess this is more a question for @rmetzger and @zentol: Is there a way to share common code between the different docker files? It looks to me that the current setup is duplicating logic across several files.

tillrohrmann (changes requested):
Thanks for creating this PR @Myasuka. I had a comment concerning the description of the newly introduced option. I think we should add this to the usage description so that users can easily learn about it.

1.11/scala_2.11-java11-debian/docker-entrypoint.sh
rmetzger: The files in the `1.11`, `1.10` and so files on `master` are generated and should not be modified manually.

`dev-master`, `dev-1.11`, `dev-1.10` contain the templates for the entry points for the different versions.

I believe we want to add this feature for upcoming releases only. I would then suggest to make the changes only on `dev-master`. As part of the release process, these changes will also end up in the `dev-1.12` branch, which we will use to generate the `1.12/` directory.

Further reading: https://github.com/apache/flink-docker#development-workflow

Myasuka:
With the default args to launch taskmanager pod, we can see the `libjemalloc` binary so is loaded in the memory process of task manager:
image

When we add `disablejemalloc` in the launch args , we can see the `libjemalloc` binary so is not loaded in the memory process of task manager:
image

BTW, to use `pmap` command in the pod, I installed the `procps` in the docker image which is not added in current PR.

1.11/scala_2.11-java11-debian/docker-entrypoint.sh
Myasuka: Thanks for @rmetzger 's explanation, I'll re-create PR for `dev-master` branch.

rmetzger: Thank you & sorry for the inconvenience.

PR #43, opened 2020-11-11
[FLINK-19125] Adopt Jemalloc as default memory allocator to avoid memory leak

Myasuka:
This PR install `libjemalloc-dev` within docker image and offer option `disablejemalloc` to start daemon. 
By default, Flink image will adopt jemalloc as default memory allocator and will disable it if option `disablejemalloc` is given.

With the default args to launch taskmanager pod, we can see the `libjemalloc` binary so is loaded in the memory process of task manager:
image

When we add `disablejemalloc` in the launch args , we can see the `libjemalloc` binary so is not loaded in the memory process of task manager:
image

BTW, to use `pmap` command in the pod to show whether `jemalloc` is enabled, I installed the `procps` in the docker image which is not added in current PR.


docker-entrypoint.sh
```diff
@@ -93,21 +94,49 @@ prepare_job_manager_start() {
...
+    printf "    Or $(basename "$0") help\n\n"
+    printf "By default, Flink image adopts jemalloc as default memory allocator and will disable jemalloc if option '${COMMAND_DISABLE_JEMALLOC}' given.\n"
     exit 0
 elif [ "$1" = "jobmanager" ]; then
     shift 1
+    disable_jemalloc_env $@
+    disabled_jemalloc="$?"
+    if [ ${disabled_jemalloc} = 0 ]; then
+      shift 1
+    fi
```
tillrohrmann, line 119: This block is repeated for all the commands. Can we simply have a function which is called with `$@` and which returns the new list of arguments for the `exec` command?

docker-entrypoint.sh
```diff
@@ -21,6 +21,7 @@
 COMMAND_STANDALONE="standalone-job"
 COMMAND_NATIVE_KUBERNETES="native-k8s"
 COMMAND_HISTORY_SERVER="history-server"
+COMMAND_DISABLE_JEMALLOC="disablejemalloc"
```
tillrohrmann, line 24: Maybe use `disable-jemalloc` instead.

tillrohrmann (commented):
Thanks for updating this PR @Myasuka. I had a question concerning the repeated code block which shifts the parameters if `disablejemalloc` was set. If this is not possible, then I am fine with accepting the PR as is.

docker-entrypoint.sh
Myasuka: I have refactored the function and update the array of arguments within the new `disable_jemalloc_env` function and then passed to next commands.

docker-entrypoint.sh
```diff
@@ -93,21 +94,41 @@ prepare_job_manager_start() {
     envsubst < "${CONF_FILE}" > "${CONF_FILE}.tmp" && mv "${CONF_FILE}.tmp" "${CONF_FILE}"
 }
 
+disable_jemalloc_env() {
+  local -n _args=$1
+  if [ "${_args[0]}" = ${COMMAND_DISABLE_JEMALLOC} ]; then
+      echo "Disable Jemalloc as the memory allocator"
+      _args=("${_args[@]:1}")
```
tillrohrmann, line 101: I am not a bash expert, hence, could you explain how this change will be visible from the outside?

tillrohrmann (commented):
Thanks for updating this PR @Myasuka. Could you please explain how this change works? Sorry I am not a bash expert.

docker-entrypoint.sh
tillrohrmann: isn't `_args` a local variable which should only be valid within the function?

Myasuka: `local -n` is the same as `declare -n` which serves as a name [reference](https://www.gnu.org/software/bash/manual/bash.html#index-declare) and I got this from [this answer](https://unix.stackexchange.com/a/449537).
This feature is introduced in [bash-4.3](https://github.com/bminor/bash/blob/ac50fbac377e32b98d2de396f016ea81e8ee9961/CHANGES#L761) and luckily our base image installed `bash-5.0.3` to include this.

tillrohrmann: ok, can we maybe add a comment to the `local -n` statement to explain the usage?

Myasuka: Sounds good, already added comment.

tillrohrmann (approved):
Thanks for updating this PR @Myasuka. LGTM. Merging this PR now.

carp84:
Somehow the commit message didn't trigger an auto-close of this PR. Since the commit already merged via b183c32, closing it manually now.

PR #46, opened 2020-12-08
[FLINK-20532] Use prope 'sed' in updated docker image + extend regex …

rmetzger:
The generated file was incorrect because the regex in the sed command didn't work as expected.
This was caused by using busybox' reduced 'sed'.

I updated the Docker image for building the stackbrew entries


zentol (approved):

generate-stackbrew-library.sh
```diff
@@ -47,7 +47,7 @@ pruneTags() {
     else
         # remove "latest" and any "scala_" tag, unless it is the latest version
```
zentol, line 48: update the comment?

rmetzger:
Thx for the review. I updated the comment & merge it now.

PR #49, opened 2020-12-17
[FLINK-20650] Mark "native-k8s" as deprecated and do the preparatory work for pass-through mode in docker-entrypoint.sh

wangyang0918:
Based on the discussion in https://github.com/docker-library/official-images/pull/9249, this PR will mark "native-k8s" as deprecated and do the preparatory work for pass-through mode in docker-entrypoint.sh. Also the `bash -c` wrapper will be removed. Then for native K8s integration, Flink should generate a full command with `bash -c` wrapper.

Then pass-through mode could be used as followings.
```
docker run flink:1.12.0 ls /opt/flink
docker run -v /path/of/kubeconfig:/opt/flink/.kube/config flink:1.12.0 bash -c 'java -classpath $FLINK_CLASSPATH -Dlog4j.configurationFile=file:/opt/flink/conf/log4j-console.properties org.apache.flink.kubernetes.entrypoint.KubernetesSessionClusterEntrypoint -Dkubernetes.cluster-id=k8s-native-session-1'
```

### DIFF
--- a/docker-entrypoint.sh
+++ b/docker-entrypoint.sh
@@ -19,6 +19,7 @@
 ###############################################################################
 
 COMMAND_STANDALONE="standalone-job"
+# Deprecated, should be remove in Flink release 1.13
 COMMAND_NATIVE_KUBERNETES="native-k8s"
 COMMAND_HISTORY_SERVER="history-server"
 COMMAND_DISABLE_JEMALLOC="disable-jemalloc"
@@ -108,7 +109,7 @@ disable_jemalloc_env() {
 
 args=("$@")
 if [ "$1" = "help" ]; then
-    printf "Usage: $(basename "$0") (jobmanager|${COMMAND_STANDALONE}|taskmanager|${COMMAND_NATIVE_KUBERNETES}|${COMMAND_HISTORY_SERVER}) [${COMMAND_DISABLE_JEMALLOC}]\n"
+    printf "Usage: $(basename "$0") (jobmanager|${COMMAND_STANDALONE}|taskmanager|${COMMAND_HISTORY_SERVER}) [${COMMAND_DISABLE_JEMALLOC}]\n"
     printf "    Or $(basename "$0") help\n\n"
     printf "By default, Flink image adopts jemalloc as default memory allocator and will disable jemalloc if option '${COMMAND_DISABLE_JEMALLOC}' given.\n"
     exit 0
@@ -171,4 +172,16 @@ elif [ "$1" = "$COMMAND_NATIVE_KUBERNETES" ]; then
     exec $(drop_privs_cmd) bash -c "${args[@]}"
 fi
 
-exec "$@"
+args=("${args[@]}")
+
+disable_jemalloc_env args
+
+copy_plugins_if_required
+
+# Set the Flink related environments
+export _FLINK_HOME_DETERMINED=true
+. $FLINK_HOME/bin/config.sh
+export FLINK_CLASSPATH="`constructFlinkClassPath`:$INTERNAL_HADOOP_CLASSPATHS"
+
+echo "Running command in pass-through mode: ${args[@]}"
+exec $(drop_privs_cmd) "${args[@]}"


wangyang0918:
cc @rmetzger @zentol 

@yosifkit do you think it makes sense now?

docker-entrypoint.sh
```diff
@@ -19,7 +19,7 @@
 ###############################################################################
 
 COMMAND_STANDALONE="standalone-job"
-COMMAND_NATIVE_KUBERNETES="native-k8s"
+COMMAND_GENERIC="generic"
```
tillrohrmann, line 22: Maybe `cmd` or `run` instead of `generic`?

tillrohrmann (approved):
Thanks for creating this PR @wangyang0918. The changes look good to me. I had only a single question about the command name.

Which changes to we have to make on the Flink side?

docker-entrypoint.sh
wangyang0918: `run` makes more sense to me. I will rename it to `run`.

wangyang0918:
@tillrohrmann Currently, in the Flink implementation, we are always passing the `native-k8s` as one of the arguments of JobManager container[1]. So when it changes in the flink-docker project, we need to update it accordingly.

[1]. https://github.com/apache/flink/blob/master/flink-kubernetes/src/main/java/org/apache/flink/kubernetes/kubeclient/decorators/JavaCmdJobManagerDecorator.java#L64

tillrohrmann:
Could you link the issue here once created @wangyang0918?

wangyang0918:
@tillrohrmann Yeah. I will attach the Flink PR here once created.

Moreover, I am not whether we should wait for the response of docker guys before merging this PR so that we will not block at image publishing next time.

tillrohrmann:
Is there already an image out there which supports the `native-k8s` command?

wangyang0918:
@tillrohrmann No, we do not have an image in the docker hub which supports `native-k8s` command. Because this PR https://github.com/docker-library/official-images/pull/9249 has not been merged.

tillrohrmann:
I see. Do we want to support this? I guess people would be confused if there is breaking API between bug fix versions. Maybe we can still support the `native-k8s` just for this release so that the `1.12.0` release keeps working.

wangyang0918:
@tillrohrmann Do you mean we still support `native-k8s` command and mark it as `Deprecated`? Then both 1.12.0 and 1.12.1+ Flink client could work with the images. We could completely remove the `native-k8s` command in 1.13.

After this, maybe we need to create a new PR for the docker publishing to replace docker-library/official-images#9249.

tillrohrmann:
Yes, this would be my suggestion in order to not break things within the `1.12.x` release line.

wangyang0918:
@tillrohrmann I have updated this PR. Once this PR is merged, we could ping the docker guys for merging the docker publishing PR for 1.12.0.

tianon:
`docker run flink run bash` just to get a shell that has useful environment variables is a bit convoluted/awkward -- are these variables harmful for normal commands?  Given that a shell still has to be used in order to actually use the variables, why not just use `bash`, or better yet, just set the variables unilaterally for all non-"commands" so that `docker run flink bash -c '...'` will just DTRT?

wangyang0918:
@tianon Thanks for your comments. We not only set the environments in `run` command, but also could support "disable jemalloc", "copy plugins" and other preparatory works in the future. For normal commands, we do not need this.

Maybe I do not give a good example in the PR description. Actually, we do not always need  a shell script to use these variables. For example, we could have the following typical commands. 
```
docker run flink:1.12.0-debug run flink run -h
docker run flink:1.12.0-debug run java -classpath '$FLINK_CLASSPATH' ... org.apache.flink.kubernetes.entrypoint.KubernetesApplicationClusterEntrypoint ...
```

tianon:
Well, now I'm confused -- `docker run flink:xxx run java -classpath '$FLINK_CLASSPATH' ...` shouldn't work at all, because nothing will expand `$FLINK_CLASSPATH`, so the classpath will be set to literally that string. :confused:

(Also, `docker run flink:xxx run flink run ...` is even _worse_ from a visual usability perspective :weary:)

rogaha:
I agree with @tianon that this approach doesn't seem to be the correct place/way to expose this functionality. Creating a specific script for running these commands or doing it through `bash` directly seem to be cleaner. thoughts @wangyang0918 ?

wangyang0918:
@rogaha Do you mean to completely remove the `run` command and moving the preparatory works(e.g. copy plugins, set environment variables, etc.) directly to work for all normal commands?

wangyang0918:
cc @rogaha @tianon I have updated this PR and prepared a testing docker image with the new entrypoint. You could directly run the following examples.

Note that if you want to run the second command you need to have a valid kube config since it means to start a native K8s session cluster.

```
docker run registry.cn-beijing.aliyuncs.com/streamcompute/flink:1.12.0-debug ls /flink
docker run -v /path/of/kubeconfig:/opt/flink/.kube/config registry.cn-beijing.aliyuncs.com/streamcompute/flink:1.12.0-debug bash -c 'java -classpath $FLINK_CLASSPATH -Dlog4j.configurationFile=file:/opt/flink/conf/log4j-console.properties org.apache.flink.kubernetes.entrypoint.KubernetesSessionClusterEntrypoint -Dkubernetes.cluster-id=k8s-native-session-1'
```

rogaha:
@wangyang0918 LGTM, we just need update the `PR` description.

wangyang0918:
@rogaha Thanks for your following comments. I have updated the description of this PR.

xintongsong (approved):
Thanks @wangyang0918 for opening this PR, and @tillrohrmann, @tianon, @rogaha for the review.
The changes LGTM. Merging now.

xintongsong:
Merged via:
* dev-master: f716a648c7b5a106b7294c22d76225bcf0edcdf3
* dev-1.12: f8b7af6fd4977c0d3e012d473838cd69c890e334

PR #54, opened 2021-01-13
[FLINK-20915][docker] Move docker entrypoint to distribution

zentol:
Counterpart to https://github.com/apache/flink/pull/14630.

CI will naturally fail on this branch since the Flink snapshots don't contain the necessary changes yet. Please refer to the CI run in the linked PR.


docker-entrypoint.sh
```diff
@@ -41,60 +31,6 @@ drop_privs_cmd() {
...
-
-    if [ -n "${FLINK_PROPERTIES}" ]; then
-        echo "${FLINK_PROPERTIES}" >> "${CONF_FILE}"
-    fi
-    envsubst < "${CONF_FILE}" > "${CONF_FILE}.tmp" && mv "${CONF_FILE}.tmp" "${CONF_FILE}"
-}
-
 disable_jemalloc_env() {
   # use nameref '_args' to update the passed 'args' within function
   local -n _args=$1
```
rmetzger, line 36: COMMAND_DISABLE_JEMALLOC is not defined in this file (anymore)

docker-entrypoint.sh
```diff
@@ -107,80 +43,10 @@ disable_jemalloc_env() {
...
 disable_jemalloc_env args
 
-copy_plugins_if_required
-
-# Set the Flink related environments
-export _FLINK_HOME_DETERMINED=true
-. $FLINK_HOME/bin/config.sh
-export FLINK_CLASSPATH="`constructFlinkClassPath`:$INTERNAL_HADOOP_CLASSPATHS"
-
 # Running command in pass-through mode
```
rmetzger, line 51:
```suggestion
# Running command in pass-through mode: The actual docker-entrypoint.sh logic is in the Flink distribution.
```

rmetzger (changes requested):
I'll comment in the other PR.

docker-entrypoint.sh
```diff
@@ -41,60 +33,6 @@ drop_privs_cmd() {
...
-    copy_plugins_if_required
-
-    set_common_options
-
-    if [ -n "${FLINK_PROPERTIES}" ]; then
-        echo "${FLINK_PROPERTIES}" >> "${CONF_FILE}"
-    fi
-    envsubst < "${CONF_FILE}" > "${CONF_FILE}.tmp" && mv "${CONF_FILE}.tmp" "${CONF_FILE}"
-}
-
```
rmetzger, line 97: @wangyang0918 mentioned that he'd prefer controlling jemalloc via an environment variable

zentol: Out-of-scope of this PR as far as I'm concerned. That's an API change that I would move to a follow-up.

wangyang0918: I believe setting the env(`export DISABLE_JEMALLOC=true`) is more convenient to use and could be easily integrated in all modes(docker, standalone, native-k8s). I agree that this could be done as a follow-up.

rmetzger: Okay, let's do it later then

rmetzger (approved):
Thanks for addressing this.

zentol:
I have updated the PR. The previous version broke the jemalloc switch since we were checking the wrong argument index. I did not want to deal with mutating arrays, and instead refactored the switch to a environment variable and added tests for the behavior.

PR #57, opened 2021-01-19
[FLINK-21034] Refactor jemalloc switch to environment variable

zentol:
Spliced out of #54.

### DIFF
--- a/docker-entrypoint.sh
+++ b/docker-entrypoint.sh
@@ -22,7 +22,6 @@ COMMAND_STANDALONE="standalone-job"
 # Deprecated, should be remove in Flink release 1.13
 COMMAND_NATIVE_KUBERNETES="native-k8s"
 COMMAND_HISTORY_SERVER="history-server"
-COMMAND_DISABLE_JEMALLOC="disable-jemalloc"
 
 # If unspecified, the hostname of the container is taken as the JobManager address
 JOB_MANAGER_RPC_ADDRESS=${JOB_MANAGER_RPC_ADDRESS:-$(hostname -f)}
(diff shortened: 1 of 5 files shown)


docker-entrypoint.sh
```diff
@@ -95,40 +94,34 @@ prepare_job_manager_start() {
...
-disable_jemalloc_env() {
-  # use nameref '_args' to update the passed 'args' within function
-  local -n _args=$1
-  if [ "${_args[0]}" = ${COMMAND_DISABLE_JEMALLOC} ]; then
-      echo "Disable Jemalloc as the memory allocator"
-      _args=("${_args[@]:1}")
-  else
-      export LD_PRELOAD=$LD_PRELOAD:/usr/lib/x86_64-linux-gnu/libjemalloc.so
-  fi
+maybe_enable_jemalloc() {
```
wangyang0918, line 97: Since the `jemalloc` is enabled by default, maybe the name `disable_jemalloc_if_required` is better. It is aligned with `copy_plugins_if_required`.

docker-entrypoint.sh
```diff
@@ -95,40 +94,34 @@ prepare_job_manager_start() {
...
+    if ! [ "${DISABLE_JEMALLOC:-false}" == "true" ]; then
```
wangyang0918, line 98:
```suggestion
    if [ "${DISABLE_JEMALLOC:-false}" != "true" ]; then
```
Or
```suggestion
    if [ "${DISABLE_JEMALLOC:-false}" == "false" ]; then
```

testing/bin/docker-entrypoint.sh
```diff
@@ -0,0 +1,37 @@
...
+if ! [ "$1" == "hello" ] || ! [ "$2" == "world" ]; then
+    echo "Incorrect arguments passed through:" "$@"
+    exit 1
+fi
+
+originalLdPreloadSetting=$3
+jemallocDisabled=$4
+
+if [ "$jemallocDisabled" == "true" ] && ! [ "$originalLdPreloadSetting" == "$LD_PRELOAD" ]; then
+    echo "jemalloc was not disabled; exploed LD_PRELOAD to be '$originalLdPreloadSetting' but was '$LD_PRELOAD'"
```
wangyang0918, line 30: typo here? `exploed` -> `expected`

wangyang0918 (commented):
I like the environment mechanism to disable jemalloc and just left some minor comments.

testing/bin/config.sh
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/env bash
```
Myasuka, line 1: This file is missing apache license.

Myasuka (commented):
Apart from this PR, we should also change [doc for flink docker](https://ci.apache.org/projects/flink/flink-docs-stable/deployment/resource-providers/standalone/docker.html#switch-memory-allocator).

docker-entrypoint.sh
zentol: I disagree. The default in the image is glibc, and we explicitly _enable_ jemalloc unless some condition is met.

wangyang0918: Make sense.

wangyang0918:
+1 for merging.

PR #58, opened 2021-01-20
[FLINK-21035] Deduplicate 'copy_plugins_if_required' calls 

zentol:
Based on #57.

PR #59, opened 2021-01-20
[FLINK-21037] Deduplicate configuration logic 

zentol:
Based on #58.

One item that specifically needs a discussion is whether it is safe to also set the number of task slots for jobmanager processes. I would generally think so, as any user not setting it themselves may run into consistency issues anyway. (relevant: FLINK-21036)

### DIFF
--- a/docker-entrypoint.sh
+++ b/docker-entrypoint.sh
@@ -74,16 +74,13 @@ set_config_option() {
   fi
 }
 
-set_common_options() {
+prepare_configuration() {
     set_config_option jobmanager.rpc.address ${JOB_MANAGER_RPC_ADDRESS}
     set_config_option blob.server.port 6124
     set_config_option query.server.port 6125
-}
 
-prepare_job_manager_start() {
-    echo "Starting Job Manager"
-
-    set_common_options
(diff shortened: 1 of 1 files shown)


docker-entrypoint.sh
```diff
@@ -112,41 +111,26 @@ if [ "$1" = "help" ]; then
...
     exec $(drop_privs_cmd) "$FLINK_HOME/bin/standalone-job.sh" start-foreground "${args[@]}"
 elif [ "$1" = ${COMMAND_HISTORY_SERVER} ]; then
     args=("${args[@]:1}")
 
     echo "Starting History Server"
 
-    if [ -n "${FLINK_PROPERTIES}" ]; then
-        echo "${FLINK_PROPERTIES}" >> "${CONF_FILE}"
-    fi
-    envsubst < "${CONF_FILE}" > "${CONF_FILE}.tmp" && mv "${CONF_FILE}.tmp" "${CONF_FILE}"
```
wangyang0918, line 139: So we do not need to do the substitution after then. Why?

zentol: I think you're being tricked by the diff; it is at the end of prepare_configuration.

wangyang0918: My bad. You are right.

wangyang0918 (approved):
+1 for merging.

PR #63, opened 2021-01-26
use full key for robert

rmetzger:


add-version.sh
```diff
@@ -112,9 +112,9 @@ for source_variant in "${SOURCE_VARIANTS[@]}"; do
 
             flink_url_file_path=flink/flink-${flink_version}/flink-${flink_version}-bin-scala_${scala_version}.tgz
 
-            flink_tgz_url="https://www.apache.org/dyn/closer.cgi?action=download&filename=${flink_url_file_path}"
+            flink_tgz_url="https://archive.apache.org/dist/${flink_url_file_path}"
```
zentol, line 115: Maybe we can make this change only locally when updating the `master` contents; we'd otherwise have to revert it pretty much immediately.

rmetzger: Why do we have to revert it?
Isn't it generally more reliable to download from the archive, than the mirrors of the current versions?

zentol: It is highly frowned upon to download anything related to recent releases from the archive.

Now, for 1.12.0 this would be _fine_, but for 1.12.1 it should still work against the mirrors.

That said, I'm not sure what the behavior would be a year down the line. Is that what you're alluding to?

zentol: In any case this has been the behavior since 1.9 and we haven't run into issues.

rmetzger: The only note I found about this, is on this page: http://archive.apache.org/, and it states
> Do note that a daily limit of 5GB per IP is being enforced on archive.apache.org, to prevent abuse.

Even though our e2e tests are downloading from archive.apache.org regularly already, we don't seem to run into any limits:
```
flink-end-to-end-tests/flink-end-to-end-tests-common-kafka/src/main/java/org/apache/flink/tests/util/kafka/LocalStandaloneKafkaResource.java:                "https://archive.apache.org/dist/kafka/%s/kafka_2.11-%s.tgz",
flink-end-to-end-tests/flink-end-to-end-tests-hbase/src/main/java/org/apache/flink/tests/util/hbase/LocalStandaloneHBaseResource.java:                "https://archive.apache.org/dist/hbase/%1$s/hbase-%1$s-bin.tar.gz", hbaseVersion);
flink-end-to-end-tests/test-scripts/docker-hadoop-secure-cluster/Dockerfile:ENV HADOOP_URL http://archive.apache.org/dist/hadoop/common/hadoop-$HADOOP_VERSION/hadoop-$HADOOP_VERSION.tar.gz
flink-end-to-end-tests/test-scripts/kafka-common.sh:  KAFKA_URL="https://archive.apache.org/dist/kafka/$KAFKA_VERSION/kafka_2.12-$KAFKA_VERSION.tgz"
```

Dockerfiles generated from this script will be build during e2e tests, and once by DockerHub when we do a new release. I believe this is an acceptable download frequency.

I'm happy to adjust this if we run into limits with archive.a.o, but for now I don't think it's worth our time overoptimizing this.

zentol: But why change it in the first place? Have we seen issues anywhere with downloads not working?

PR #67, opened 2021-03-03
Dev 1.12: Add GPG and update testing script for 1.12.2

rkhachatryan:


rmetzger:
I manually restarted the travis build

rmetzger:
It seems there's something wrong with the `wget -nv -O flink.tgz https://www.apache.org/dyn/closer.cgi?action=download&filename=flink/flink-1.12.2/flink-1.12.2-bin-scala_2.11.tgz` command. Also locally, I'm facing issues with the wget call (downloading from the browser works).

rkhachatryan:
You are right. Quotes are missing in url.

rmetzger:
Why is this error surfacing now? The Dockerfile hasn't been updated? https://github.com/apache/flink-docker/blame/dev-master/Dockerfile-debian.template#L61

rkhachatryan:
I don't see any changes in the scripts generating URLs too at least since Jul 2020.
Probably the base image changed?

I opened a PR to check if quoting fixes the problem: https://github.com/apache/flink-docker/pull/70

zentol:
Given that other builds are currently passing without the added quotes I'd think that this is just mirrors causing problems.

rmetzger:
Maybe the 1.12.2 release hasn't been synched to all mirrors yet?
Shall we use archive.apache.org 🙄  ?

zentol:
the weird thing is that if the mirrors weren't synced wget would just fail and not stall. It looks more like we're hitting an exceptionally slow mirro :/

rmetzger:
@rkhachatryan maybe push a version to travis where wget is verbosely logging what's going on?

zentol:
> Shall we use archive.apache.org

Given the issues that we're struggling with in the Flink repo I don't think this is a viable option.

zentol:
Issues seems to be resolved itself.

zentol (approved):
+1; please remember to revert the wget change

PR #74, opened 2021-04-29
Updated maintainers

AHeise:


generate-stackbrew-library.sh
```diff
@@ -33,13 +33,14 @@ source common.sh
 cat <<-EOH
 # this file is generated via https://github.com/apache/flink-docker/blob/$(fileCommit "$self")/$self
 
-Maintainers: Patrick Lucas <me@patricklucas.com> (@patricklucas),
+Maintainers: Robert Metzger <rmetzger@apache.org> (@rmetzger),
+             Chesnay Schepler <chesnay@apache.org> (@chesnay),
```
rmetzger, line 37: @zentol ok with you?

rmetzger (approved):
+1 to merge from my point of view. (let's wait for @zentol's approval as well)

zentol:
Do we need to list specific people, or can we just point to the Flink project as a whole?

AHeise:
Having the project is probably more future proof. Not sure why you need to list maintainers here in the first place...

zentol:
> Not sure why you need to list maintainers here in the first place...

Because the docker folks require it: https://github.com/docker-library/official-images#instruction-format

I've found a case where this field [just contains the project](https://github.com/docker-library/official-images/blob/master/library/express-gateway), so let's do that as well.

AHeise:
> > Not sure why you need to list maintainers here in the first place...
> 
> Because the docker folks require it: https://github.com/docker-library/official-images#instruction-format
> 
> I've found a case where this field [just contains the project](https://github.com/docker-library/official-images/blob/master/library/express-gateway), so let's do that as well.

Could you please add a suggestion?

generate-stackbrew-library.sh
rmetzger, line 37:
```suggestion
Maintainers: The Apache Flink Project <dev@flink.apache.org> (@ApacheFlink)
```

generate-stackbrew-library.sh
```diff
@@ -33,13 +33,14 @@ source common.sh
...
              Ismaël Mejía <iemejia@gmail.com> (@iemejia)
 GitRepo: https://github.com/apache/flink-docker.git
```
rmetzger, line 39:
```suggestion
Maintainers: The Apache Flink Project <dev@flink.apache.org> (@ApacheFlink)
GitRepo: https://github.com/apache/flink-docker.git
```

generate-stackbrew-library.sh
```diff
@@ -33,13 +33,12 @@ source common.sh
...
 
-Maintainers: Patrick Lucas <me@patricklucas.com> (@patricklucas),
-             Ismaël Mejía <iemejia@gmail.com> (@iemejia)
+Maintainers: The Apache Flink Project <dev@flink.apache.org> (@ApacheFlink)
 GitRepo: https://github.com/apache/flink-docker.git
 EOH
 
 
-for dockerfile in $(find . -name "Dockerfile"); do
+for dockerfile in $(find . -name "Dockerfile" | sort -r); do
```
zentol, line 41: can you move this into a separate commit?

zentol (approved):
+1, with a small commit massaging

AHeise:
Extracted that into a separate commit and merging.

PR #80, opened 2021-05-12
[FLINK-22648] Only set number of slots if env variable is set

zentol:
Small modification to the entrypoint to prevent surprising modifications to the number of slots.


fapaul (approved):
Thanks for adding this fix. I verified that if the variable is not set by the entry point or an external source the default value is still `1`. In summary AFAIS this does not introduce unexpected behaviour 👍

PR #83, opened 2021-05-28
Release 1.13.1

dawidwys:


testing/run_travis_tests.sh
```diff
@@ -11,12 +11,9 @@ fi
...
 BRANCH="$TRAVIS_BRANCH"
 
-test_docker_entrypoint
-
-./add-custom.sh -u "https://s3.amazonaws.com/flink-nightly/flink-1.13-SNAPSHOT-bin-scala_2.11.tgz" -n test-java8
+./add-version.sh -r 1.13 -f 1.13.1
 
-# test Flink with Java11 image as well
-./add-custom.sh -u "https://s3.amazonaws.com/flink-nightly/flink-1.13-SNAPSHOT-bin-scala_2.11.tgz" -j 11 -n test-java11
+test_docker_entrypoint
```
AHeise, line 16: This looks odd to me. I think this should be in a different commit that explains what you intended.

AHeise (commented):
This doesn't look like a simple version to me.

testing/run_travis_tests.sh
dawidwys: Previously we run the tests against the snapshot version of Flink because 1.13.0 was not released yet. I forgot to update the file once 1.13.0 was released.

See the `dev-1.12` branch

AHeise: Okay understood.

AHeise (approved):

PR #88, opened 2021-08-03
Release 1.12.5

JingsongLi:


xintongsong (changes requested):
We also need to update gpg key for the new version in add-version.sh

JingsongLi:
@xintongsong Thanks for the review, gpg key is here: https://github.com/apache/flink-docker/commit/a71b798f908c92e5164a0985d15a4f76ca745a3b

xintongsong (approved):
LGTM

PR #94, opened 2021-12-10
[FLINK-25251] Use Java 11 by default

zentol:


XComp (commented):
`add-custom.sh:15` specifies the Java version to be `8` if nothing is specified. I feel like we should rely on `DEFAULT_JAVA` there as well:

The following code code be added below the parameter while loop in `add-custom.sh`:
```

if [ -z "${java_version}" ]; then
  [ -z "${DEFAULT_JAVA}" ] && echo "Neither -j nor DEFAULT_JAVA are specified." && exit 1
  java_version="${DEFAULT_JAVA}"
fi
```

add-custom.sh
```diff
@@ -12,7 +12,7 @@ function usage() {
 
 binary_download_url=
 name=custom
-java_version=8
+java_version=11
```
XComp, line 15: Still, I would have preferred to rely on the `DEFAULT_JAVA` environment variable which is exported in `generator.sh` to have only one location that defines the default Java version.

XComp (approved):
I have a minor concern. See my comment below. But the code works. 👍

add-custom.sh
```diff
@@ -12,7 +12,7 @@ function usage() {
 
 binary_download_url=
 name=custom
-java_version=8
+java_version=${DEFAULT_JAVA}
```
XComp, line 15: Moving this initialization below the while loop using the following code would make the script more robust:
```
if [ -z "${java_version}" ]; then
  [ -z "${DEFAULT_JAVA}" ] && echo "Neither -j nor DEFAULT_JAVA are specified." && exit 1
  java_version="${DEFAULT_JAVA}"
fi
```

zentol: Shouldn't we rather fail in this very line if DEFAULT_JAVA is not set? And that you could achieved with `-u`.

PR #99, opened 2022-01-21
[FLINK-25679] Add support for arm64v8

rmetzger:


rmetzger:
Thanks for your review!

PR #111, opened 2022-04-26
[FLINK-27413] Drop Scala 2.11 support 

zentol:


add-version.sh
```diff
@@ -58,7 +58,7 @@ if [[ ! "$flink_version" =~ ^$flink_release\.+ ]]; then
 fi
 
 # Defaults, can vary between versions
-scala_versions=( 2.11 2.12 )
+scala_versions=( 2.12 )
 java_versions=( 8 11 )
 gpg_key=
```
XComp, line 64: Why isn't it necessary to fix the `scala_versions` for specific Flink versions but the GPG key?

testing/run_travis_tests.sh
```diff
@@ -13,10 +13,10 @@ BRANCH="$TRAVIS_BRANCH"
 
 test_docker_entrypoint
 
-./add-custom.sh -u "https://s3.amazonaws.com/flink-nightly/flink-1.15-SNAPSHOT-bin-scala_2.11.tgz" -j 8 -n test-java8
+./add-custom.sh -u "https://s3.amazonaws.com/flink-nightly/flink-1.16-SNAPSHOT-bin-scala_2.12.tgz" -j 8 -n test-java8
```
XComp, line 16: Is it worth maintaining this file considering that we're not using travis anymore? 🤔 I couldn't find a usage of the script anywhere...

zentol: flink-docker uses travis for CI.

add-version.sh
zentol: because the same key is used to sign both releases.

XComp: I guess, I have some misconception how this script is used. I would have assumed setting some fixed configuration for specific version (i.e. the GPG key), we should also fix the scala version for these versions. 🤔

zentol: In practice the gpg keys are only required in the dev-1.X branches, because thats where we generate the actual docker files from.

zentol: technically we also don't need this list of keys, but just a single key per branch.

zentol: in the 1.15 branch the key is there: https://github.com/apache/flink-docker/commit/8d95fddb0bcee059c69665bffacda2dbbfc14d1a

zentol: so essentially, we don't need to add the 1.15 key to dev-master because dev-1.15 already exists, and the key is irrelevant for future releases (i.e., dev-1.16)

XComp: I see, thanks for clarification.

XComp (approved):
LGTM 👍 The only other appearance of `2.11` is in the docker-test-job which seems to be ok because we're using out-dated Flink 1.9, anyway and just using basic API inside the job

PR #112, opened 2022-04-26
[1.15][FLINK-27413] Drop Scala 2.11 support

zentol:


XComp (approved):
LGTM The only other appearance of 2.11 is in the `docker-test-job` which seems to be ok because we're using out-dated Flink 1.9, anyway and just using basic API inside the job

PR #116, opened 2022-05-06
[FLINK-27671][ci] Publish snapshot images to GHCR

afedulov:


zentol (changes requested):
I've spotted a fair number of flaws. It would be good to clarify which files you expect to live on which branch. When it comes to GHA this is quite important; you can't trigger push/pr builds for the dev-branches from the master branch, nor cron builds from the dev-branches. Re-using workflows that are not self-contained (i.e., rely on some scripts) can also be tricky because those files are not pulled in automatically.

.github/workflows/snapshot.yml
```diff
@@ -0,0 +1,68 @@
...
+    branches:
+      - snapshot-dev-master
+
+env:
+  REGISTRY: ghcr.io
+  VERSION: 1.16-SNAPSHOT
+
+jobs:
+  ci:
+    uses: ./.github/workflows/ci.yml
```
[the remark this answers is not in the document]
afedulov: @zentol I did not know about this limitation. Do you propose to skip the CI step?

afedulov (commented):
@zentol Thanks for the review. I addressed your comments. There are still some remaining questions regarding the branches/scheduled workflow setup.

utils.sh
```diff
@@ -0,0 +1,59 @@
...
+    local image_tag image_name dockerfile_dir
+    image_tag="$(image_tag "$dockerfile")"
+    image_name="$(image_name "$image_tag")"
+    dockerfile_dir="$(dirname "$dockerfile")"
+
+    echo >&2 "===> Building ${image_tag} image..."
+    docker build -t "$image_name" "$dockerfile_dir"
+}
+
+function publish_snapshots() {
```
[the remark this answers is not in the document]
afedulov: I guess you are right, I'll inline this logic into the pipeline.

afedulov:
@zentol I cleaned up the PR and resolved conflicts with the master, PTAL.
The main question for me is where this change should be merged. IIUC the issue is that the add-custom.sh script is not available on master, while scheduled runs can only be executed there. Would you be OK with adding add-custom.sh to master  in that case?

.github/workflows/snapshot.yml
```diff
@@ -0,0 +1,71 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+name: "Publish SNAPSHOTs"
+
+on:
+  # TODO: enable before merge
+  # schedule:
+  #   - cron: '0 0 * * *' # Deploy every day
```
MartijnVisser, line 21: Reminder to enable this

.github/workflows/snapshot.yml
```diff
@@ -0,0 +1,71 @@
...
+  push:
+    branches:
+      - snapshot-dev-master
```
MartijnVisser, line 24: Should this be removed? We only want this workflow to run on schedule or manually triggered, right?

MartijnVisser (approved):
LGTM overall

zentol:
> Would you be OK with adding add-custom.sh to master in that case?

No; you can't add add-custom.sh to master because it's ultimately just a utility wrapper of the docker file generation scripts. we would have to move all of that to master, but that breaks the development model of flink-docker. It would also lock us out of creating snapshot images for other releases (e.g., 1.15).

> I thought that we were going to put the snapshot.yaml to the dev-master branch. If I get you correctly, this won't work with the scheduled runs, because they can only be triggered from master. Is my understanding correct?

Correct; the workflow file for the scheduled run must be defined in the master branch.

However, you _can_ checkout other branches.

I mentioned this before, but https://github.com/apache/flink/blob/master/.github/workflows/docs.yml is really good example on how to achieve this. You have the workflow defined on master, but it checks out the dev branches and uses the scripts from those branches to do stuff. You can think of the scripts on the dev branches to be a sort of interface that the scheduled run works against. This does imply that we _may_ duplicate some stuff from the workflow file (i.e., if we want to run ci), and we _may_ have issues when the scripts are being changed across versions. But that happens rarely enough that I'm not too worried about it.

zentol:
Note that right now I'd avoid CI as a whole because there are some conceptual issues. Specifically, the tests build their own images, so technically we aren't testing what we release.

.github/workflows/snapshot.yml
afedulov: Yes, this is here just to be able to see the results in GHCR. Will be removed pre-merge after the PR approval.

.github/workflows/snapshot.yml
```diff
@@ -0,0 +1,71 @@
...
+      - name: Prepare Dockerfiles
+        run: |
+          ./add-custom.sh -u "https://s3.amazonaws.com/flink-nightly/flink-${{ matrix.flink_version }}-bin-scala_2.12.tgz" -j ${{ matrix.java_version }} -n ${IMAGE_NAME}
+      - name: Build image
+        run: |
+          dockerfile="$(ls ./*/*${{ matrix.flink_version }}*${{ matrix.java_version }}*/Dockerfile)"
+          dockerfile_dir="$(dirname "$dockerfile")"
+          tag="${REGISTRY}/${OWNER}/${IMAGE_REPO}:${IMAGE_NAME}-debian"
+          echo "===> Building ${tag} image from ${dockerfile_dir}"
+          docker build -t "${REGISTRY}/${OWNER}/${IMAGE_REPO}:${IMAGE_NAME}-debian" "${dockerfile_dir}"
```
alpreu, line 64: Could use the tag variable

.github/workflows/snapshot.yml
```diff
@@ -0,0 +1,71 @@
...
+      - name: Docker login
+        run: |
+          docker login ${REGISTRY} -u ${{ github.actor }} -p ${{ secrets.GITHUB_TOKEN }}
+      - name: "Publish snapshots"
+        run: |
+          docker push ${REGISTRY}/${OWNER}/${IMAGE_REPO}:${IMAGE_NAME}-debian
```
gaoyunhaii, line 70: Here we should also use the shared `${tags}`. We may move creating the variable to `Set env` step

afedulov: Which additional tags do you think we should add?

.github/workflows/snapshot.yml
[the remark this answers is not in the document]
afedulov: Disabled CI.

afedulov:
Thanks everyone for the input.
@zentol I adjusted according to the proposed branches checkout approach, PTAL.

zentol:
Manually merged.

PR #117, opened 2022-06-14
[FLINK-28057] Fix LD_PRELOAD on ARM images

nferrario:
This fixes an error that prevents jemalloc from being loaded on ARM images:
```
ERROR: ld.so: object '/usr/lib/x86_64-linux-gnu/libjemalloc.so' from LD_PRELOAD cannot be preloaded (cannot open shared object file): ignored.
```

### DIFF
--- a/docker-entrypoint.sh
+++ b/docker-entrypoint.sh
@@ -91,7 +91,20 @@ prepare_configuration() {
 
 maybe_enable_jemalloc() {
     if [ "${DISABLE_JEMALLOC:-false}" == "false" ]; then
-        export LD_PRELOAD=$LD_PRELOAD:/usr/lib/x86_64-linux-gnu/libjemalloc.so
+        JEMALLOC_PATH="/usr/lib/$(uname -m)-linux-gnu/libjemalloc.so"
+        JEMALLOC_FALLBACK="/usr/lib/x86_64-linux-gnu/libjemalloc.so"
+        if [ -f "$JEMALLOC_PATH" ]; then
+            export LD_PRELOAD=$LD_PRELOAD:$JEMALLOC_PATH
+        elif [ -f "$JEMALLOC_FALLBACK" ]; then
+            export LD_PRELOAD=$LD_PRELOAD:$JEMALLOC_FALLBACK
+        else
+            if [ "$JEMALLOC_PATH" = "$JEMALLOC_FALLBACK" ]; then
+                MSG_PATH=$JEMALLOC_PATH
+            else
+                MSG_PATH="$JEMALLOC_PATH and $JEMALLOC_FALLBACK"
+            fi
+            echo "WARNING: attempted to load jemalloc from $MSG_PATH but the library couldn't be found. glibc will be used instead."
+        fi
     fi
 }
 
--- a/testing/bin/docker-entrypoint.sh
+++ b/testing/bin/docker-entrypoint.sh
@@ -27,11 +27,19 @@ originalLdPreloadSetting=$3
 jemallocDisabled=$4
 
 if [ "$jemallocDisabled" == "true" ] && ! [ "$originalLdPreloadSetting" == "$LD_PRELOAD" ]; then
-    echo "jemalloc was not disabled; expected LD_PRELOAD to be '$originalLdPreloadSetting' but was '$LD_PRELOAD'"
+    echo "jemalloc was disabled; expected LD_PRELOAD to be '$originalLdPreloadSetting' but was '$LD_PRELOAD'"
     exit 1
 fi
 
+# We expect jemalloc to be available in any of these paths
+JEMALLOC_PATH="/usr/lib/$(uname -m)-linux-gnu/libjemalloc.so"
+JEMALLOC_FALLBACK="/usr/lib/x86_64-linux-gnu/libjemalloc.so"
+
 if [ "$jemallocDisabled" == "false" ] && [ "$originalLdPreloadSetting" == "$LD_PRELOAD" ]; then
-    echo "jemalloc was disabled; expected LD_PRELOAD to be different than '$originalLdPreloadSetting'."
+    if [ ! -f "$JEMALLOC_PATH" ] && [ ! -f "$JEMALLOC_FALLBACK" ]; then
+      echo "jemalloc was enabled but it was not found in the system. LD_PRELOAD is unchanged and glibc will be used instead."
+      exit 0
+    fi
+    echo "jemalloc was enabled; expected LD_PRELOAD to be different than '$originalLdPreloadSetting'."
     exit 1
 fi


1.15/scala_2.12-java11-debian/docker-entrypoint.sh
```diff
@@ -91,7 +91,12 @@ prepare_configuration() {
 
 maybe_enable_jemalloc() {
     if [ "${DISABLE_JEMALLOC:-false}" == "false" ]; then
-        export LD_PRELOAD=$LD_PRELOAD:/usr/lib/x86_64-linux-gnu/libjemalloc.so
+        # Maybe use export LD_PRELOAD=$LD_PRELOAD:/usr/lib/$(uname -i)-linux-gnu/libjemalloc.so
+        if [[ `uname -i` == 'aarch64' ]]; then
+            export LD_PRELOAD=$LD_PRELOAD:/usr/lib/aarch64-linux-gnu/libjemalloc.so
```
Myasuka, line 96: We need a `Dockerfile` could be built on arm platform successfully then the change in `docker-entrypoint.sh` is reasonable.

Myasuka (commented):
As I said in the issue ticket, do we need to update the `Dockerfile` to work with this change?

nferrario:
> These changes need to be applied to the dev-master/dev-1.15 branch. The master branch only contains generated content.

Will fix, thanks

1.15/scala_2.12-java11-debian/docker-entrypoint.sh
Myasuka: I see. 
Will `uname -i` always return `aarch64` on ARM platform?

Myasuka:
@nferrario actually, you should make this PR target for `dev-master` branch first, and we can cherry-pick the final commit to `dev-1.15` after code review if no obvious code conflict.

1.15/scala_2.12-java11-debian/docker-entrypoint.sh
zentol: If that is a concern, would it be a crazy idea to just check if the file exists?

That seems like a safer approach anyway.

nferrario: Since we're using a Debian-based image, `uname` should always return what we expect. The reason why I said `in case uname -i returns something other than x86_64 on an actual x64 host` is to NOT affect x64, not even by accident.

We can definitely check whether the file exists. I feel like `uname -i` is a cleaner approach, but I'm open to suggestions :)

nferrario:
> @nferrario actually, you should make this PR target for `dev-master` branch first, and we can cherry-pick the final commit to `dev-1.15` after code review if no obvious code conflict.

Done!

docker-entrypoint.sh
```diff
@@ -91,7 +91,12 @@ prepare_configuration() {
 
 maybe_enable_jemalloc() {
     if [ "${DISABLE_JEMALLOC:-false}" == "false" ]; then
-        export LD_PRELOAD=$LD_PRELOAD:/usr/lib/x86_64-linux-gnu/libjemalloc.so
+        # Maybe use export LD_PRELOAD=$LD_PRELOAD:/usr/lib/$(uname -i)-linux-gnu/libjemalloc.so
+        if [[ `uname -i` == 'aarch64' ]]; then
+            export LD_PRELOAD=$LD_PRELOAD:/usr/lib/aarch64-linux-gnu/libjemalloc.so
```
KarlManong, line 96: why not using `$(uname -m)` directly in the line?

such as 
```
export LD_PRELOAD=$LD_PRELOAD:/usr/lib/$(uname -m)-linux-gnu/libjemalloc.so
```

Myasuka: From the description of [uname](https://www.gnu.org/software/coreutils/manual/html_node/uname-invocation.html#uname-invocation), we can see `uname -m` is a better choice.

1.15/scala_2.12-java11-debian/docker-entrypoint.sh
Myasuka: Based on current comments, we can use `uname -m` to get the machine hardware name, and check whether the file existed. If not, we fall back to default location `/usr/lib/x86_64-linux-gnu/libjemalloc.so`.

docker-entrypoint.sh
nferrario: You're absolutely right, it is `-m` indeed. Fixing now. Thank you all for your input

1.15/scala_2.12-java11-debian/docker-entrypoint.sh
Myasuka: @nferrario did you ever consider this suggestion?

nferrario: Yes I just made the changes and read this comment last, so I'm adding the fallback to that directory, although it looks like uname -m will never return something invalid

Myasuka: Thanks for the update. I said we can fall back to the default location, however, we still need to check whether the default location file existed. Otherwise, we need to print a warning message to tell users that even if we want to enable JEMALLOC, the dependency library does not exist.

nferrario: Considering we already get the error I posted in the first PR comment as the very first line of the Job Manager and Task Manager docker logs, Idk how much more it would help to add more warnings. It already is a very low level thing that you won't be able to notice unless you specifically look for it, because there's no way to know that Jemalloc isn't being used (it took months for us to realize).
I will however add a message that's easier to understand, since the default LD_PRELOAD error still requires the user to look at the source code or the Docker image structure at the very least.

nferrario: @Myasuka  Hmm, the tests pass if I run `bash testing/run_tests.sh` instead of `./testing/run_tests.sh`. Is it ok if I change `ci.yml` to:
```yaml
- name: "Test images"
   run: bash testing/run_tests.sh
```

nferrario: @Myasuka I don't think there's a good way to have the tests actually check if LD_PRELOAD is set. I'm not sure if you'd want me to `touch`, which might be fine on Github Actions but might be annoying when testing locally, or just ignore the test entirely.

Myasuka: > It already is a very low level thing that you won't be able to notice unless you specifically look for it, because there's no way to know that Jemalloc isn't being used (it took months for us to realize).

Actually, you can use `pmap <pid>` to check whether the process has attached the required library.

I think this PR looks good without unit test as the logic is quite easy, and it's a bit complex to verify this behavior.

Myasuka (approved):
@nferrario Could you squash all commits into one single commit?

This PR LGTM.
@zentol do you have more comments?

1.15/scala_2.12-java11-debian/docker-entrypoint.sh
nferrario: True, but for that we need to containerize the testing suite so that we can include those libs, and then have cross-platform testing with something like `docker buildx` (or mock those paths). I commented out the test and added a link to this thread.

Myasuka: It makes me confused about why we need to disable `testing/testing_lib.sh` if in the current status.

nferrario: Because LD_PRELOAD used to have a value even when jemalloc didn't exist in the host. Now LD_PRELOAD will be empty on that scenario, and since the test suite isn't running on a Flink Docker image, but rather on just plain Github (or a local machine), we only have glibc available

nferrario: I agree, but the test will never work unless we run a Dockerized test. We cannot fake a system library in a way that works for multiple OS (Linux, macOS)

Myasuka: @nferrario We can adjust the test itself instead of just disabling the test, e.g. we can directly `exit 0` if the expected jemalloc binary so file does not exist.

nferrario: @Myasuka I just added that to the test. It's still useless until we move this to a Docker image, but we at least won't need to make a code change when we do so.

Myasuka:
@nferrario would you please create PR targets for dev-1.14 and dev-1.15 branch?

PR #153, opened 2023-03-23
[FLINK-31566][docker] Removes 1.15 and adds 1.17 Dockerfiles

XComp:
This also includes some minor documentation fixes where I removed steps that are not necessary anymore (see the commit message for the reasoning)

### DIFF
--- a/.github/workflows/snapshot.yml
+++ b/.github/workflows/snapshot.yml
@@ -38,8 +38,10 @@ jobs:
       matrix:
         java_version: [8, 11]
         build:
-          - flink_version: 1.17-SNAPSHOT
+          - flink_version: 1.18-SNAPSHOT
             branch: dev-master
+          - flink_version: 1.17-SNAPSHOT
+            branch: dev-1.17
           - flink_version: 1.16-SNAPSHOT
             branch: dev-1.16
           - flink_version: 1.15-SNAPSHOT
--- a/1.17/scala_2.12-java11-ubuntu/Dockerfile
+++ b/1.17/scala_2.12-java11-ubuntu/Dockerfile
@@ -1,0 +1,94 @@
+###############################################################################
+#  Licensed to the Apache Software Foundation (ASF) under one
+#  or more contributor license agreements.  See the NOTICE file
+#  distributed with this work for additional information
+#  regarding copyright ownership.  The ASF licenses this file
+#  to you under the Apache License, Version 2.0 (the
+#  "License"); you may not use this file except in compliance
+#  with the License.  You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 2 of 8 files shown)


XComp:
I needed to do a force-push because I wanted to update the commit message for the commit that added the 1.17.0 Dockerimages to comply with the README.

XComp:
And another force push was necessary because I realized that we had the GHA workflows not covered.

.github/workflows/snapshot.yml
```diff
@@ -38,12 +38,12 @@ jobs:
...
         java_version: [8, 11]
         build:
-          - flink_version: 1.17-SNAPSHOT
+          - flink_version: 1.18-SNAPSHOT
             branch: dev-master
+          - flink_version: 1.17-SNAPSHOT
+            branch: dev-1.17
           - flink_version: 1.16-SNAPSHOT
             branch: dev-1.16
-          - flink_version: 1.15-SNAPSHOT
```
XComp, line 45: tbh. here I'm not 100% sure considering [the discussion](https://lists.apache.org/thread/szq23kr3rlkm80rw7k9n95js5vqpsnbv) we had about having a final patch version release.

PatrickRen: According to the discussion we still need to release 1.15.5 as the final flushing version. Then I think 1.15 related  contents should be kept until the final 1.15.5 is released. Otherwise we will never be able to release the 1.15.5 to the docker-library/official-images repo, as a new release requires to point to a commit on master. 

IIUC the workflow handling an end-of-support version (let say 1.15 here) would be:
- We keep 1.15 related contents on master
- Once the final version 1.15.5 is ready, release the Docker image following the guide in README.md
- Remove 1.15 related contents from master after 1.15.5 is successfully published to Docker official
- Remove 1.15 related contents from [library/flink](https://github.com/docker-library/official-images/blob/master/library/flink) in official-images repo

WDYT?

PatrickRen (commented):
Thanks for the PR, and detailed updates on the README @XComp ! This truly helps release managers in future releases. 

I do have some concerns considering the discussion about final patch version you mentioned. I replied in your comments.

.github/workflows/snapshot.yml
XComp: Initially, I had the concern that we would miss deleting it if the community decide that they don't want another patch release for the deprecated minor version (for the current release I mean 1.15.5). But you're right - in the worst case we delete it later on.

rmetzger (approved):
Changes look good. Thanks for updating the docs.

XComp:
Thanks for the review, @PatrickRen and @rmetzger . I squashed the changes into meaningful chunks (and left 1.15 data in for now). @rmetzger will pick up the work on FLINK-31566 after the changes are merged.

PR #158, opened 2023-06-05
[ADD-SYMLINKS] Add symlinks for FLINK_VERSION to FLINK_RELEASE jars

daniel-packard:
Hi flink team - we're using the `flink:1.16` base image to deploy many of our flink jobs, but because this image is built on top of a more specific bugfix release (1.16.X), we run into issues where referencing specific jar versions in our deployment configurations (e.g. `flink-libary-1.16.2.jar`) break when the underlying base image bumps the bugfix version number.

This PR provides a fix wherein the FLINK_RELEASE (1.16) is symlinked to the underlying FLINK_VERSION (1.16.2) jars.

```
# e.g. 
> ls opt
....
 flink-python-1.16.jar -> flink-python-1.16.2.jar
....
``` 

This is a similar pattern to the way many linux distros handle sxs installations of versioned libraries. 

In our deployments, we would now use `flink:1.16` docker base image, _and_ reference jars by the `1.16.jar` version, instead of being forced to reference them by a specific bugfix version that occasionally breaks

NOTE: we're aware that we can _also_ use the `flink:1.16.2` base image - but this locks us out of bugfix releases. 

If you like this pattern, I'd be happy to port to the other dev branches as well

### DIFF
--- a/generator.sh
+++ b/generator.sh
@@ -8,12 +8,17 @@ export DEFAULT_JAVA="11"
 function generateDockerfile {
     # define variables
     dir=$1
-    binary_download_url=$2
-    asc_download_url=$3
+    flink_release=$2
+    flink_version=$3
     gpg_key=$4
     check_gpg=$5
     java_version=$6
     source_variant=$7
+
+    flink_url_file_path=flink/flink-${flink_version}/flink-${flink_version}-bin-scala_${scala_version}.tgz
+    flink_tgz_url="https://www.apache.org/dyn/closer.cgi?action=download&filename=${flink_url_file_path}"
+    # Not all mirrors have the .asc files
+    flink_asc_url=https://www.apache.org/dist/${flink_url_file_path}.asc
 
     from_docker_image="eclipse-temurin:${java_version}-jre-jammy"
 
@@ -24,6 +29,9 @@ function generateDockerfile {
 
     # generate Dockerfile
     sed \
+        -e "s,%%FLINK_RELEASE%%,${flink_release}," \
+        -e "s,%%FLINK_VERSION%%,${flink_version}," \
+        -e "s,%%BINARY_DOWNLOAD_URL%%,${escaped_binary_download_url}," \
         -e "s,%%BINARY_DOWNLOAD_URL%%,${escaped_binary_download_url}," \
         -e "s,%%ASC_DOWNLOAD_URL%%,$asc_download_url," \
         -e "s/%%GPG_KEY%%/$gpg_key/" \
--- a/generator.sh
+++ b/generator.sh
@@ -13,7 +13,8 @@ function generateDockerfile {
     gpg_key=$4
     check_gpg=$5
     java_version=$6
-    source_variant=$7
+    scala_version=$6
+    source_variant=$8
 
     flink_url_file_path=flink/flink-${flink_version}/flink-${flink_version}-bin-scala_${scala_version}.tgz
     binary_download_url="https://www.apache.org/dyn/closer.cgi?action=download&filename=${flink_url_file_path}"
--- a/generator.sh
+++ b/generator.sh
@@ -13,7 +13,7 @@ function generateDockerfile {
     gpg_key=$4
     check_gpg=$5
     java_version=$6
-    scala_version=$6
+    scala_version=$7
     source_variant=$8
 
     flink_url_file_path=flink/flink-${flink_version}/flink-${flink_version}-bin-scala_${scala_version}.tgz
--- a/generator.sh
+++ b/generator.sh
@@ -16,9 +16,9 @@ function generateDockerfile {
     source_variant=$7
 
     flink_url_file_path=flink/flink-${flink_version}/flink-${flink_version}-bin-scala_${scala_version}.tgz
-    flink_tgz_url="https://www.apache.org/dyn/closer.cgi?action=download&filename=${flink_url_file_path}"
+    binary_download_url="https://www.apache.org/dyn/closer.cgi?action=download&filename=${flink_url_file_path}"
     # Not all mirrors have the .asc files
-    flink_asc_url=https://www.apache.org/dist/${flink_url_file_path}.asc
+    asc_download_url=https://www.apache.org/dist/${flink_url_file_path}.asc
 
     from_docker_image="eclipse-temurin:${java_version}-jre-jammy"
 
--- a/Dockerfile-ubuntu.template
+++ b/Dockerfile-ubuntu.template
@@ -44,7 +44,9 @@ RUN set -ex; \
   gosu nobody true
 
 # Configure Flink version
-ENV FLINK_TGZ_URL=%%BINARY_DOWNLOAD_URL%% \
+ENV FLINK_RELEASE=%%FLINK_RELEASE%% \
+    FLINK_VERSION=%%FLINK_VERSION%% \
+    FLINK_TGZ_URL=%%BINARY_DOWNLOAD_URL%% \
     FLINK_ASC_URL=%%ASC_DOWNLOAD_URL%% \
     GPG_KEY=%%GPG_KEY%% \
     CHECK_GPG=%%CHECK_GPG%%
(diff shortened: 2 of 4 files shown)


daniel-packard:
PS - I would also love some guidance for testing these changes to ensure I'm not breaking anything

MartijnVisser:
I hope that @zentol would have the room to help out here

generator.sh
```diff
@@ -8,12 +8,18 @@ export DEFAULT_JAVA="11"
...
     dir=$1
-    binary_download_url=$2
-    asc_download_url=$3
+    flink_release=$2
+    flink_version=$3
     gpg_key=$4
     check_gpg=$5
     java_version=$6
-    source_variant=$7
+    scala_version=$6
```
zentol, line 16:
```suggestion
    scala_version=$7
```

zentol (commented):
Wouldn't globbing patterns or similar allow you to handle this, like `flink-s3-fs-presto-1.16.*.jar`?

I was sometimes wondering why we include versions in the jar names in the first place, but this would have to be changed in Flink itself.

> NOTE: we're aware that we can also use the flink:1.16.2 base image - but this locks us out of bugfix releases.

You'd have to explicitly upgrade to the next Flink patch version, but that may be a good idea anyway depending on how you deploy Flink.
You do _not_ want different processes of Flink using different patch versions, so unless you have safeguards to prevent that in the event that you deploy something at the exact time we do a release, you should maybe consider it.

daniel-packard:
> You'd have to explicitly upgrade to the next Flink patch version, but that may be a good idea anyway depending on how you deploy Flink. You do _not_ want different processes of Flink using different patch versions, so unless you have safeguards to prevent that in the event that you deploy something at the exact time we do a release, you should maybe consider it.

I'm not sure I follow your intent here... are you saying that `1.16.1` -> `1.16.2` is an unsafe version transition to happen automatically?

daniel-packard:
> Wouldn't globbing patterns or similar allow you to handle this, like flink-s3-fs-presto-1.16.*.jar?

I didn't know this was an option in helm/k8s config - but I will look into it to see if it's possible!

daniel-packard:
> Wouldn't globbing patterns or similar allow you to handle this, like flink-s3-fs-presto-1.16.*.jar?

quick followup here - i think this would only work if the flink kubernetes operator supported file globbing on these config properties: 

```yaml
apiVersion: flink.apache.org/v1beta1
kind: FlinkDeployment
spec:
  ...
  flinkConfiguration:
    ...
    containerized.master.env.ENABLE_BUILT_IN_PLUGINS: "flink-s3-fs-presto-1.16.1.jar"
    containerized.taskmanager.env.ENABLE_BUILT_IN_PLUGINS: "flink-s3-fs-presto-1.16.1.jar"
   ....
```

... and I don't think it supports this - they must be exact file names

zentol:
> Or are you saying that it's unsafe to deploy mixed flink patch/bugfix versions (e.g. some references to 1.16.1 & other references to 1.16.2 in the same cluster)?

This is the unsafe bit I was referring to.

> I'm not sure I follow your intent here... are you saying that 1.16.1 -> 1.16.2 is an unsafe version transition to happen automatically?

This may be another unsafe bit, depending on how you deploy Flink.
Let's say you have a cluster with HA.
You can't just upgrade all the pods to a newer version, but need to stop all jobs with a savepoint, upgrade _all_ pods (== upgrade the cluster), then re-submit the jobs with their respective savepoint.

> containerized.master.env.ENABLE_BUILT_IN_PLUGINS: "flink-s3-fs-presto-1.16.1.jar"
> containerized.taskmanager.env.ENABLE_BUILT_IN_PLUGINS: "flink-s3-fs-presto-1.16.1.jar"

These shouldn't handled by the operator itself, but by the docker entrypoint:
https://github.com/apache/flink-docker/blob/master/1.17/scala_2.12-java11-ubuntu/docker-entrypoint.sh#L41

So we're back in good ol' bash-land, and it boils down to whether that method works with `*`.
At a glance, it appears to not work. `mkdir`/`ln` work fine, but the file existence checking isn't expanding `*`. So if we want to go down this route, maybe we can adjust this line to use `ls` (which totally works) instead.

daniel-packard:
@zentol - thanks for the detailed response! 

I'll defer to you on the preferred way to handle this. I'm more comfortable making the change here by adding symlinks, but I think updating `docker-entrypoint.sh` to use `ls` instead of `[ -e ... ]` also makes sense - I'm just less sure how to make that change 😅 

something like this?

```sh
ls ${FLINK_HOME}/opt/${target_plugin}

if [  $? != 0 ]; then
  # if return code of `ls <pattern>` is NON-zero, jar not found
  echo "Plugin ${target_plugin} does not exist. Exiting."
  exit 1
else
  # if return code of `ls <pattern>` is zero, jar found -> symlink it
  ln -fs "${FLINK_HOME}/opt/${target_plugin}" "${FLINK_HOME}/plugins/${plugin_name}"
  echo "Successfully enabled ${target_plugin}"
fi
```

Would there be an issue if _more than one_ file matched the regex?

zentol:
> Would there be an issue if more than one file matched the regex?

yes; that would cause multiple plugins to be copied into the same directory, likely breaking them (the directory isn't just for show; they define the classloading boundaries).

daniel-packard:
Hey @zentol - 

Right now we have two proposed changes on the table:
1. add symlinks from e.g.  `1.16.jar` -> `1.16.2.jar`
2. add support for pattern matching e.g. `1.16.*.jar`

I think we should try to decide between: 
- implement 1, 
- implement 2, 
- implement _both_, 
- implement _neither_

Both of the proposed changes would let the consumer of `flink:1.16` docker images ignore the detail of the patch version (e.g. `1.16.2`). 

So I think the first question we should answer is... Is it a reasonable request that "consumers of `flink:1.16` docker images" should be able to "ignore the detail of the patch version"?

In my mind, the answer is "yes" - because if I wanted to worry about patch version, I would use a more specific docker base image (`flink:1.16.2`) 

But it's possible that safety/stability _demands_ that devs are aware of changing patch versions .... in which case the answer would be no.. But then I might ask why `flink:1.16` is even an option for a base image if we still have to worry about patch version changes?

zentol:
> But then I might ask why flink:1.16 is even an option for a base image if we still have to worry about patch version changes?

That's an interesting line of thought. To be honest I believe the answer to that question is that we just inherited that and didn't think   much about it.

daniel-packard:
> we just inherited that and didn't think much about it

tradition is strong in software 😅

Any thoughts on which direction to go here?

At this point, I'm inclined to pin our base docker image to a patch version (e.g. `flink:1.16.2`) to avoid unexpected breaking config when `flink:1.16` gets bumped to a new patch version... although I'm still a bit dissatisfied that we would miss out on future patches/fixes unless we explicitly bump all the versions 😞

zentol:
I've started a mailing list discussion because I don't want to decide this on my own.

https://lists.apache.org/thread/5wrnqn1hhv1hxjg9fmz2c8v0qc0h4kmr

daniel-packard:
I appreciate you taking a look, and sharing with the community! 

In the meantime, we will pin our flink base image versions to the patch version (e.g. `1.16.2`)

patricklucas:
> But then I might ask why `flink:1.16` is even an option for a base image if we still have to worry about patch version changes?

Hi @daniel-packard, original author of flink-docker here.

During the process of making these images available we found that developing "official" Docker images always comes with some tradeoffs between the different ways the images are used, especially with respect to configuration defaults, for example, but also image tags.

We want to have production-ready images, of course, but also make it convenient for ad hoc usage by developers—and most Flink users typically think of the Flink minor version they are using rather than the patch version (i.e. a user should be able to `docker run flink` or `docker run flink:1.17` and get some useful output.

For production usage, it's usually a good idea to anchor yourself to a sufficiently-specific version which can vary by application (`debian:12` is probably fine; for Python you probably want at least `python:3.11`; for Flink, I would go with `flink:1.17.1`—sometimes users even pin to a specific `@sha256:...`).

So to answer your question, there's a good argument to be made that we should only publish Flink patch version tags, but that would come with the tradeoff of preventing new users of the images from running `docker run flink` or using `flink:1.17` in their Docker Compose configs, for instance.

daniel-packard:
> So to answer your question, there's a good argument to be made that we should only publish Flink patch version tags, but that would come with the tradeoff of preventing new users of the images from running docker run flink or using flink:1.17 in their Docker Compose configs, for instance.

Thanks for the response! 

IMO the tradeoff for beginners of `docker run flink` vs `docker run flink:1.17.x` is totally worth it to remove a foot gun for production deployments - but I understand the motivation.

I'm gonna close this issue, since we've resolved it for our team by pinning our flink base image to a PATCH version.

PR #170, opened 2024-01-22
[FLINK-34165] Update apache download url

JingGe:
Apache changed its download link.


MartijnVisser (changes requested):
@JingGe I checked the download sites at the ASF:

> dlcdn.apache.org is our CDN for downloads.a.o, which should arguably be faster than just going straight to downloads.a.o. While the CDN is global, we've had issues in certain countries where it's either slow or unavailable, hence the direct link. Using either is totally fine. 

I think we can point all links to https://dlcdn.apache.org. ~~I've verified that this works for existing/active supported versions (currently 1.18, 1.17 and 1.16) but also for older versions (1.15), because that will redirect straight to the archive.~~

Examples:

This all works, supported versions
https://dlcdn.apache.org/flink/flink-1.18.1/flink-1.18.1-bin-scala_2.12.tgz
https://dlcdn.apache.org/flink/flink-1.18.1/flink-1.18.1-bin-scala_2.12.tgz.asc
https://dlcdn.apache.org/flink/flink-1.18.1/flink-1.18.1-bin-scala_2.12.tgz.sha512

This doesn't work for CDN, since it's not a supported version anymore
https://dlcdn.apache.org/flink/flink-1.15.0/flink-1.15.0-bin-scala_2.12.tgz   
https://dlcdn.apache.org/flink/flink-1.15.0/flink-1.15.0-bin-scala_2.12.tgz.asc
https://dlcdn.apache.org/flink/flink-1.15.0/flink-1.15.0-bin-scala_2.12.tgz.sha512

This ~~works~~ doesn't work for all supported/non-supported releases (binary for non-supported version is broken)
https://downloads.apache.org/flink/flink-1.18.1/flink-1.18.1-bin-scala_2.12.tgz
https://downloads.apache.org/flink/flink-1.18.1/flink-1.18.1-bin-scala_2.12.tgz.asc
https://downloads.apache.org/flink/flink-1.18.1/flink-1.18.1-bin-scala_2.12.tgz.sha512
https://downloads.apache.org/flink/flink-1.15.0/flink-1.15.0-bin-scala_2.12.tgz
https://downloads.apache.org/flink/flink-1.15.0/flink-1.15.0-bin-scala_2.12.tgz.asc
https://downloads.apache.org/flink/flink-1.15.0/flink-1.15.0-bin-scala_2.12.tgz.sha512

Edit: wait, it doesn't work for https://downloads.apache.org/flink/flink-1.15.0/flink-1.15.0-bin-scala_2.12.tgz
**So, then I think we could change all of the links to dlcdn.apache.org :)**

JingGe:
@MartijnVisser yes, https://dlcdn.apache.org/ is another option which works too. I will update to use that one. Thanks!

JingGe:
BTW, there are two urls defined for tgz and asc. This PR only changes asc url. For asc, we could use dns but the difference should be very limited. WDYT?

MartijnVisser:
> For asc, we could use cdn but the difference should be very limited. WDYT?

I think we should keep them the same for consistency

JingGe:
> I think we should keep them the same for consistency

Make sense. I am thinking why it was not consistent in the first place.

JingGe:
There is a comment for it: 

`# Not all mirrors have the .asc files` 
https://github.com/apache/flink-docker/blob/df317cb99bdc28b9f0aafe8616aa075a07499fbf/add-version.sh#L128

I think we should indeed use the direct download url. WDYT?

MartijnVisser (approved):

JingGe:
Thanks @MartijnVisser for the review!

PR #175, opened 2024-01-23
[FLINK-34205] Use BashJavaUtils for Flink configuration management

JunRuiLee:
The flink-docker's Dockerfile and docker-entrypoint.sh currently use shell scripting techniques with grep and sed for configuration reading and modification. This method is not suitable for the standard YAML configuration format.

Following the changes introduced in [FLINK-33721](https://issues.apache.org/jira/browse/FLINK-33721), we should update flink-docker's Dockerfile and docker-entrypoint.sh to use BashJavaUtils for Flink configuration reading and writing.


zhuzhurk (approved):
LGTM.

PR #181, opened 2024-02-28
[FLINK-34419][docker] Added support of JDK 17 & 21 for Flink 1.18+ 

morazow:
[[FLINK-34419](https://issues.apache.org/jira/browse/FLINK-34419)]:  Adds support of JDK 17 & 21 for Flink 1.18+ versions in the snapshot builds.

### DIFF
--- a/.github/workflows/snapshot.yml
+++ b/.github/workflows/snapshot.yml
@@ -50,21 +50,18 @@ jobs:
             java_version: [8, 11]
             branch: dev-1.17
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
         with:
           ref: ${{ matrix.build.branch }}
 
       - name: Set up QEMU
-        uses: docker/setup-qemu-action@v1
-        with:
-          image: tonistiigi/binfmt:latest
-          platforms: all
+        uses: docker/setup-qemu-action@v3
 
       - name: Set up Docker Buildx
-        uses: docker/setup-buildx-action@v1
+        uses: docker/setup-buildx-action@v3
 
       - name: Log in to the Container registry
-        uses: docker/login-action@v1
+        uses: docker/login-action@v3
         with:
           registry: ${{ env.REGISTRY }}
           username: ${{ github.actor }}
@@ -82,7 +79,7 @@ jobs:
         run: env
 
       - name: Build and push Docker images (supported platforms)
-        uses: docker/bake-action@v1.7.0
+        uses: docker/bake-action@v4
         with:
           files: |
             .github/workflows/docker-bake.hcl
--- a/.github/workflows/snapshot.yml
+++ b/.github/workflows/snapshot.yml
@@ -36,15 +36,18 @@ jobs:
     strategy:
       max-parallel: 1
       matrix:
-        java_version: [8, 11]
         build:
           - flink_version: 1.20-SNAPSHOT
+            java_version: [8, 11, 17, 21]
             branch: dev-master
           - flink_version: 1.19-SNAPSHOT
+            java_version: [8, 11, 17, 21]
             branch: dev-1.19
           - flink_version: 1.18-SNAPSHOT
+            java_version: [8, 11, 17]
             branch: dev-1.18
           - flink_version: 1.17-SNAPSHOT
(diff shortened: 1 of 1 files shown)


.github/workflows/snapshot.yml
```diff
@@ -36,34 +36,34 @@ jobs:
...
       - name: Set up Docker Buildx
-        uses: docker/setup-buildx-action@v1
+        uses: docker/setup-buildx-action@v3
 
       - name: Log in to the Container registry
-        uses: docker/login-action@v1
+        uses: docker/login-action@v3
         with:
-          registry: ghcr.io
+          registry: ${{ env.REGISTRY }}
```
XComp, line 66: where is this env variable initialized?

.github/workflows/snapshot.yml
```diff
@@ -36,34 +36,34 @@ jobs:
...
             branch: dev-1.19
           - flink_version: 1.18-SNAPSHOT
+            java_version: [8, 11, 17]
             branch: dev-1.18
           - flink_version: 1.17-SNAPSHOT
+            java_version: [8, 11]
             branch: dev-1.17
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
```
XComp, line 53: Thanks for upgrading the versions as well. Could you move that change into a separate hotfix commit? You can alternatively use FLINK-34314 for the commit message.

.github/workflows/snapshot.yml
```diff
@@ -36,34 +36,34 @@ jobs:
...
         with:
           ref: ${{ matrix.build.branch }}
 
       - name: Set up QEMU
-        uses: docker/setup-qemu-action@v1
-        with:
-          image: tonistiigi/binfmt:latest
-          platforms: all
+        uses: docker/setup-qemu-action@v3
```
XComp, line 58: I [verified](https://github.com/docker/setup-qemu-action?tab=readme-ov-file#inputs) that we don't need to set these explicitly because they are the default values for these parameters

XComp (changes requested):
LGTM. :+1: I have a few minor comments PTAL.

.github/workflows/snapshot.yml
morazow: This is set in the beginning of the file, on the `env` part

.github/workflows/snapshot.yml
morazow: Done

morazow:
Thanks @XComp for the review, I have applied your suggestions, please have a look.

.github/workflows/snapshot.yml
XComp: True, that one I overlooked. :facepalm:

XComp (approved):
Thanks for addressing my comments. LGTM :+1:

morazow:
Thanks @XComp for review and approval, could you please merge them?

XComp:
I will merge everything at once. That way FLINK-34419 can be resolved in a consistent way.

.github/workflows/snapshot.yml
```diff
@@ -36,34 +36,34 @@ jobs:
...
+            java_version: [8, 11, 17, 21]
             branch: dev-master
           - flink_version: 1.19-SNAPSHOT
+            java_version: [8, 11, 17, 21]
             branch: dev-1.19
           - flink_version: 1.18-SNAPSHOT
+            java_version: [8, 11, 17]
             branch: dev-1.18
           - flink_version: 1.17-SNAPSHOT
+            java_version: [8, 11]
```
dstreitfeld-sn, line 50: @morazow @XComp this change broke the "Publish SNAPSHOTs" workflow on master.  the problem arises in the "Prepare Dockerfiles and set env" step below (lines 70-76), where there are references to `matrix.java_version` which is no longer set.  for example, see https://github.com/apache/flink-docker/actions/runs/14481889740/job/40620349446

morazow: Thanks @dstreitfeld-sn for reporting!

Indeed it is referencing wrongly. I have added PR, https://github.com/apache/flink-docker/pull/223. Please have a look 🤝

PR #11, opened 2020-04-13
[FLINK-17167] Add HistoryServer option

derjust:
Adds 'historyserver' as option to start the Docker container.
This option uses Flink's `start-historyserver.sh` start script to start a Flink history server.

PR #182, opened 2024-02-28
[FLINK-34419][docker] Add tests for JDK 17 & 21

morazow:
[[FLINK-34419](https://issues.apache.org/jira/browse/FLINK-34419)]: Updated tests to use JDK 17 and 21 versions in `dev-master`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,7 +29,7 @@ jobs:
       matrix:
         java_version: [ 8, 11, 17, 21 ]
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
       - name: "Build images"
         run: |
           ./add-custom.sh -u "$FLINK_TAR_URL" -j ${{ matrix.java_version }} -n "test-java${{ matrix.java_version }}"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,16 +17,21 @@ name: "CI"
 
 on: [push, pull_request]
 
+env:
+  FLINK_TAR_URL: "https://s3.amazonaws.com/flink-nightly/flink-1.20-SNAPSHOT-bin-scala_2.12.tgz"
+
 jobs:
   ci:
+    name: CI using JDK ${{ matrix.java_version }}
     runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        java_version: [ 8, 11, 17, 21 ]
     steps:
       - uses: actions/checkout@v3
       - name: "Build images"
         run: |
-          ./add-custom.sh -u "https://s3.amazonaws.com/flink-nightly/flink-1.20-SNAPSHOT-bin-scala_2.12.tgz" -j 8 -n test-java8
-          ./add-custom.sh -u "https://s3.amazonaws.com/flink-nightly/flink-1.20-SNAPSHOT-bin-scala_2.12.tgz" -j 11 -n test-java11
-          ./add-custom.sh -u "https://s3.amazonaws.com/flink-nightly/flink-1.20-SNAPSHOT-bin-scala_2.12.tgz" -j 17 -n test-java17
-          ./add-custom.sh -u "https://s3.amazonaws.com/flink-nightly/flink-1.20-SNAPSHOT-bin-scala_2.12.tgz" -j 21 -n test-java21
+          ./add-custom.sh -u "$FLINK_TAR_URL" -j ${{ matrix.java_version }} -n "test-java${{ matrix.java_version }}"
       - name: "Test images"
         run: testing/run_tests.sh
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -26,5 +26,7 @@ jobs:
         run: |
           ./add-custom.sh -u "https://s3.amazonaws.com/flink-nightly/flink-1.20-SNAPSHOT-bin-scala_2.12.tgz" -j 8 -n test-java8
           ./add-custom.sh -u "https://s3.amazonaws.com/flink-nightly/flink-1.20-SNAPSHOT-bin-scala_2.12.tgz" -j 11 -n test-java11
+          ./add-custom.sh -u "https://s3.amazonaws.com/flink-nightly/flink-1.20-SNAPSHOT-bin-scala_2.12.tgz" -j 17 -n test-java17
+          ./add-custom.sh -u "https://s3.amazonaws.com/flink-nightly/flink-1.20-SNAPSHOT-bin-scala_2.12.tgz" -j 21 -n test-java21
       - name: "Test images"
         run: testing/run_tests.sh


XComp (approved):

.github/workflows/ci.yml
```diff
@@ -17,18 +17,19 @@ name: "CI"
...
+env:
+  TAR_URL: "https://s3.amazonaws.com/flink-nightly/flink-1.20-SNAPSHOT-bin-scala_2.12.tgz"
+
 jobs:
   ci:
-    name: CI using JDK ${{ matrix.java_version }}
     runs-on: ubuntu-latest
-    strategy:
-      fail-fast: false
-      matrix:
```
XComp: Why did you remove the matrix? I like this approach (because it reduces the amount of redundant code to a minmum).

XComp (commented):
Switching back to not-approved to discuss the matrix approach of your first commit. :)

.github/workflows/ci.yml
morazow: Yes agree, it also isolates the tests to their own environment which is good.

There was some docker issue with matrix setup, I will update it again and check that tests run successfully.

morazow:
Thanks @XComp, I'll ping you again once I update it

morazow:
Morning @XComp,

I have updated the PR accordingly. I had to set the `max-parallel` job to 1, otherwise the occasionally the docker container names (jobmanager or taskmanager) would clash and fail to start.

I could change the test scripts to name containers so they don't clash. Please let me know.

.github/workflows/ci.yml
```diff
@@ -17,14 +17,22 @@ name: "CI"
...
+
 jobs:
   ci:
+    name: CI using JDK ${{ matrix.java_version }}
     runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      max-parallel: 1
+      matrix:
+        java_version: [ 8, 11, 17, 21 ]
```
XComp, line 31: Can you separate the commit:
- [hotfix][ci] Updates actions
- [hotfix][ci] Utilize matrix strategy in GHA workflow
- [FLINK-34419][docker] Adds JDK17 and JDK21 support

.github/workflows/ci.yml
```diff
@@ -17,14 +17,22 @@ name: "CI"
...
+env:
+  TAR_URL: "https://s3.amazonaws.com/flink-nightly/flink-1.20-SNAPSHOT-bin-scala_2.12.tgz"
+
 jobs:
   ci:
+    name: CI using JDK ${{ matrix.java_version }}
     runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      max-parallel: 1
```
XComp, line 29: > I could change the test scripts to name containers so they don't clash. Please let me know.
Sorry, I cannot follow. Where does the name clash? :thinking: Can you elaborate a bit more? 

And if it's actually like that, we might want to add a comment here that explains why `max-parallel: 1` is used.

XComp (changes requested):

.github/workflows/ci.yml
morazow: Fixed! I also changed accordingly other branches

.github/workflows/ci.yml
morazow: Done

XComp: Thanks, but I still don't understand. It seems like I'm missing something here. In what sense can name clashes appear here? The builds are executed on separate machines and we are building Docker images here? :thinking:

morazow: Yes, I am also trying to understand it. My initial thoughts were based on this failed action: https://github.com/morazow/flink-docker/actions/runs/8081181748/job/22079149028

But let me internet search and read more, maybe there are some other parameter we are missing. I would also expect that those docker containers are created in separated env for each matrix 👀

morazow:
Another off-topic point, the tests build and package a job using `mvn package` command. This step pulls many transitive dependencies. Should we add GitHub action to cache the mvn jars in this issue? Or should we address it separately?

.github/workflows/ci.yml
morazow: But the rerun of the job is fine: https://github.com/morazow/flink-docker/actions/runs/8081181748

XComp: > #11 8.399 + /bin/bash /opt/flink/bin/config-parser-utils.sh /opt/flink/conf /opt/flink/bin /opt/flink/lib -repKV rest.address,localhost,0.0.0.0 -repKV rest.bind-address,localhost,0.0.0.0 -repKV jobmanager.bind-host,localhost,0.0.0.0 -repKV taskmanager.bind-host,localhost,0.0.0.0 -rmKV taskmanager.host=localhost
> #11 8.625 sed: can't read /config.yaml: No such file or directory

Maybe, there's something odd with the config modification logic (see [workflow log](https://github.com/morazow/flink-docker/actions/runs/8081181748/job/22079149028#step:4:8838))?

morazow: Read about the GHA concurrency: https://docs.github.com/en/actions/using-jobs/using-concurrency#using-concurrency-in-different-scenarios

My assumption is that I pushed another commit while another workflow was running and maybe they clashed on container name.

I would suggest to add concurrency definition to our matrix builds:

```
concurrency:
      group: ${{ github.workflow }}-${{ github.ref }}-java-${{ matrix.java_version }}
      cancel-in-progress: true
```

This way if commits pushed one after the other only the last commit will run the workflow.

morazow: > > #11 8.399 + /bin/bash /opt/flink/bin/config-parser-utils.sh /opt/flink/conf /opt/flink/bin /opt/flink/lib -repKV rest.address,localhost,0.0.0.0 -repKV rest.bind-address,localhost,0.0.0.0 -repKV jobmanager.bind-host,localhost,0.0.0.0 -repKV taskmanager.bind-host,localhost,0.0.0.0 -rmKV taskmanager.host=localhost
> > #11 8.625 sed: can't read /config.yaml: No such file or directory
> 
> I forgot that we're also doing testing. Maybe, there's something odd with the config modification logic (see [workflow log](https://github.com/morazow/flink-docker/actions/runs/8081181748/job/22079149028#step:4:8838))?

Yes, tests have these kind of issues. Another ones:

```
WARNING: attempted to load jemalloc from /usr/lib/x86_64-linux-gnu/libjemalloc.so but the library couldn't be found. glibc will be used instead.
./docker-entrypoint.sh: line 80: /home/runner/work/flink-docker/flink-docker/testing/bin/config-parser-utils.sh: No such file or directory
jemalloc was enabled but it was not found in the system. LD_PRELOAD is unchanged and glibc will be used instead.
./docker-entrypoint.sh: line [8](https://github.com/morazow/flink-docker/actions/runs/8096433961/job/22125972877#step:4:9)0: /home/runner/work/flink-docker/flink-docker/testing/bin/config-parser-utils.sh: No such file or directory
```

But that line still happens when it is green also. Here the success run: https://github.com/morazow/flink-docker/actions/runs/8096433961/job/22125972520

XComp: > My assumption is that I pushed another commit while another workflow was running and maybe they clashed on container name. But I tried to trigger several workflow runs, still could not reproduce the issue above.

But my understanding is that jobs run on separate VMs which means that even if two jobs run in parallel starting a container with the same name would mean that they don't have any effect on each other. But I might be wrong here. :thinking: 

Can't we test that easily just starting a playground GHA workflow with two jobs where both start a container with the same name. If your claim is correct, we should see the same error appearing there. WDYT?

morazow: That is also what I think, they are run on isolated VMs.

I already restarted multiple CI runs [`run1`](https://github.com/morazow/flink-docker/actions/runs/8096425906), [`run2`](https://github.com/morazow/flink-docker/actions/runs/8096433961) on my fork. These should create four containers with same name. But so far, I could not reproduce the initial issue.

I didn't want to introduce flaky tests, going forward I would suggest to remove the set 1 parallelism. We could address it again if we see similar issue.

morazow: Okay run with `cleanup` (removes docker related resources) disabled in the tests, now there are issues: https://github.com/morazow/flink-docker/actions/runs/8112296736/job/22173221898. But this is expected.

The initial issue may be was result of these two test contentions:

```
smoke_test_all_images
smoke_test_one_image_non_root
```

Both build image, create docker network, and run test. So maybe there was an issue cleanup from one test to next test within a job.

Maybe we could refactor to cleanup method, run it before each test start with idempotent checks.

https://github.com/morazow/flink-docker/blob/dev-master/testing/testing_lib.sh#L202

morazow: I would suggest to change the testing script, I will update with another commit.

With matrix strategy, we'll have only one docker file, these scripts check all the docker files.

https://github.com/morazow/flink-docker/blob/dev-master/testing/testing_lib.sh#L226-L242

XComp: You could try to run the tests in a loop in CI to see whether you can reproduce the issue. So far, I couldn't come up with a reason for the test failure, yet. :thinking:

morazow: Thanks @XComp,

I have added loop test, and indeed everything worked, the issue didn't happen (CI run: https://github.com/morazow/flink-docker/actions/runs/8409563442/job/23027006083).

I think we could safely have matrix setup without parallelism set to `1`.

morazow:
@XComp please have a look again. If this looks good, I will backport to other branches accordingly, thanks 🙏

XComp (approved):
Looks good to me. :+1: Thanks. Can you please merge the parallelism hotfix commit into matrix hotfix commit? There's no reason to keep these two separate.

morazow:
Thanks @XComp for looking into 🙏 I have squashed it as you suggested, fixed.

I will update the other PRs similar to this one tomorrow. Thanks a lot!

XComp:
Ok, ping me when you're done updating PR #183 and PR #184

morazow:
Thanks @XComp, I have updated both pull requests

PR #186, opened 2024-03-15
Update Dockerfiles for 1.19.0 release

lincoln-lil:
* Update Dockerfiles for 1.19.0 release

### DIFF
--- a/1.19/scala_2.12-java11-ubuntu/Dockerfile
+++ b/1.19/scala_2.12-java11-ubuntu/Dockerfile
@@ -1,0 +1,105 @@
+###############################################################################
+#  Licensed to the Apache Software Foundation (ASF) under one
+#  or more contributor license agreements.  See the NOTICE file
+#  distributed with this work for additional information
+#  regarding copyright ownership.  The ASF licenses this file
+#  to you under the Apache License, Version 2.0 (the
+#  "License"); you may not use this file except in compliance
+#  with the License.  You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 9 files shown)


1.19/scala_2.12-java11-ubuntu/Dockerfile
```diff
@@ -0,0 +1,105 @@
...
+    rm -rf "$GNUPGHOME" flink.tgz.asc; \
+  fi; \
+  \
+  tar -xf flink.tgz --strip-components=1; \
+  rm flink.tgz; \
+  \
+  chown -R flink:flink .; \
+  \
+  # Replace default REST/RPC endpoint bind address to use the container's network interface \
+  CONF_FILE="$FLINK_HOME/conf/flink-conf.yaml"; \
```
MartijnVisser, line 84: Is this still correct, given the new YAML parser?

MartijnVisser (approved):
I think it's good, left one comment to double check

1.19/scala_2.12-java11-ubuntu/Dockerfile
lincoln-lil: Good catch! I checked both the FLINK-33297 and [FLIP-366](https://cwiki.apache.org/confluence/display/FLINK/FLIP-366%3A+Support+standard+YAML+for+FLINK+configuration),  the default configuration file has been changed to `config.yaml`. Also verified the dist package, we should use the new `config.yaml` instead.

lincoln-lil: I checked both the FLINK-33297 and FLIP-366, the default configuration file has been changed to `config.yaml`, but the script is compatible with the old `flink-conf.yaml` file. 
This is FLIP-366's expected behavior (For compatibility reasons, in Flink 1.x, if the old configuration file `flink-conf.yaml` exists in the Flink conf directory, Flink will ignore the new configuration file and use the old parser to parse `flink-conf.yaml` as the Flink configuration. In Flink 2.x, Flink will no longer support parsing the old configuration file `flink-conf.yaml`).

lincoln-lil (commented):
Thanks @MartijnVisser for reviewing this!

PR #197, opened 2024-06-17
Update GPG key for 1.19.1 release to full key ID

hlteoh37:
Update GPG key for 1.19.1 release to full key ID

Verified that lookup on pgp works
https://keyserver.ubuntu.com/pks/lookup?search=6378E37EB3AAEA188B9CB0D396C2914BB78A5EA1&fingerprint=on&op=index


mbalassi (approved):
Thanks, verified manually.

PR #202, opened 2024-08-23
[FLINK-33677][core] Remove flink-conf.yaml from flink dist.

JunRuiLee:
Since FLINK-33677 will remove flink-conf.yaml from flink dist. So this PR will accordingly remove the flink-conf support.


zhuzhurk (approved):
LGTM.

PR #203, opened 2024-08-23
[FLINK-35584][docker] Support Java21 in v1.20.0

JEnglandOhalo:
Add docker files for that Java 21 can be used with version 1.20.0 of Apache Flink.

### DIFF
--- a/1.20/scala_2.12-java21-ubuntu/Dockerfile
+++ b/1.20/scala_2.12-java21-ubuntu/Dockerfile
@@ -1,0 +1,105 @@
+###############################################################################
+#  Licensed to the Apache Software Foundation (ASF) under one
+#  or more contributor license agreements.  See the NOTICE file
+#  distributed with this work for additional information
+#  regarding copyright ownership.  The ASF licenses this file
+#  to you under the Apache License, Version 2.0 (the
+#  "License"); you may not use this file except in compliance
+#  with the License.  You may obtain a copy of the License at
+#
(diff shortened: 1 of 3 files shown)


priyankasunilnairtrimble:
@reswqa I noticed that most of the recent commits were done by you. Are there any plans to support this change any time soon so that we have a docker image for Flink 1.20 that supports Java 21?

reswqa:
I didn't learn about any plans for 1.20 to support jdk21, or maybe I missed some discussion.

cc @snuyanzin as you've done most of the work for jdk21 support.

JEnglandOhalo:
> I didn't learn about any plans for 1.20 to support jdk21, or maybe I missed some discussion.
> 
> cc @snuyanzin as you've done most of the work for jdk21 support.

I understand that Flink 1.19 added beta support for Java 21 - see https://www.confluent.io/en-gb/blog/exploring-apache-flink-1-19/

But, unfortunately, no OCI containers were released which meant we couldn't integrate with it in our codebase and so had to stay on Java 17.

snuyanzin:
we had all the tests cron jobs for jdk21 since 1.19
the last bit (for development) is enablement of spotless done in 2.0 since this task required dropping of jdk 1.8

At the same time there is still an activity FLINK-33699 which finally was decided not to do however still. makes sense to keep in mind FLINK-33699.
There is no tests for  savepoints stored with jdk 1.8/11/17 and restored in jdk21

PR #204, opened 2024-09-24
[FLINK-36360] Fix version name checks for preview release.

xintongsong:


add-version.sh
```diff
@@ -53,7 +53,7 @@ if [ -z "$flink_release" ] || [ -z "$flink_version" ]; then
     exit 1
 fi
 
-if [[ ! "$flink_version" =~ ^$flink_release\.+ ]]; then
+if [[ "$flink_version" != "$flink_release"* ]]; then
```
davidradl, line 56: the way it is written, this will accept 1.2 as both the Flink version and release, but the example in the above comments is  
```
flink_release= # Like 1.2
flink_version= # Like 1.2.0
```
implies that the version is always longer than the release. This looks inconsistent.

xintongsong: We are preparing for a 2.0-preview release, where:
```
flink_release="2.0"
flink_version="2.0-preview1"
```
This is currently rejected by the current check.

Moreover, these scripts are mainly used for creating docker images for Flink releases. The release managers know what version names to expect, and are not highly relying on these checks.

reswqa (approved):

PR #217, opened 2025-03-06
[FLINK-32482][docker] Add Java 21 to build matrix

reswqa:


add-version.sh
```diff
@@ -59,7 +59,7 @@ fi
 
 # Defaults, can vary between versions
 scala_versions=( 2.12 )
-java_versions=( 8 11 17 )
+java_versions=( 11 17 21 )
```
snuyanzin, line 62: is the same script use for 1.20.x?

if yes, then jdk8 still might be required for 1.20.2
Similar question bout default

snuyanzin (approved):

reswqa:
@snuyanzin The CI not pickup for jdk11 because we push nightly binary release to s3 with jdk17(Cannot run under jdk11 setup)... I think we should change it also to jdk11(https://github.com/apache/flink/pull/26260).

PR #225, opened 2025-06-01
FLINK-37881: Drop gosu in favour of Dockerfile USER

avi-sanwal:
This is a minor improvement from security standpoint on flink's docker image.

In the dockerfile of flink's docker image, we are adding gosu:
https://github.com/apache/flink-docker/blob/6e226503dbb228467905c70ccfb6f33f4c676872/1.20/scala_2.12-java17-ubuntu/Dockerfile#L27-L44
This is later used to switch to the user flink in the entrypoint script: https://github.com/apache/flink-docker/blob/6e226503dbb228467905c70ccfb6f33f4c676872/1.20/scala_2.12-java17-ubuntu/docker-entrypoint.sh#L37.

Gosu itself is tagged by popular scanners as being vulnerable due to outdated golang usage (though gosu itself is not).

Instead of using gosu/su-exec, it may be preferrable to switch to flink user using USER flink or something similar in Dockerfile. This way we could avoid depending on gosu, and present a cleaner scan result.

While this does not really solve any vulnerabilties in the system, it can help satisfy auditors against false positive reports provided by heuristical scanners.


avi-sanwal:
> In general I agree to remove `gosu` but I suggest to do some testing because docker files are user facing. An immediate what the user will recognize that docker inheritance + install something would not work. That would mean all users must adjust their custom docker files. For large organizations that can be heavy or impossible so we must consider and highlight what this change would mean.
> 
> I think this PR must go into the `dev-master` branch generator script, when it's there the generated materials can be copied here. Can you do it plz?

Thanks for your time reviewing this change. I will do this via `dev-master`.
Agreed, it may break compatibility.

I have already done this in our own base image with a modified copy of the entrypoint script and it has been running successfully in production for a few months now.

We use flink kubernetes operator managing the deployments.

PR #227, opened 2025-07-22
[FLINK-38010][release] Update dev-master point to 2.2

lsyldliu:


lincoln-lil:
Could you have a check if we should also upgrade the flink version in testing/docker-test-job/pom.xml ?

lincoln-lil:
Thanks for confirming!  +1

PR #246, opened 2025-12-03
Update Dockerfiles for 2.2.0 release

ruanhang1993:


.github/workflows/docker_push.yml
```diff
@@ -73,6 +73,16 @@ jobs:
           - flink_version: "2.1"
             java_version: "21"
             dockerfile: "2.1/scala_2.12-java21-ubuntu/Dockerfile"
+          # Flink 2.2.x (Latest)
```
lincoln-lil, line 76: Since we added 2.2, should the 2.0.x be removed?

lincoln-lil: We can discuss it in a new thread.

lincoln-lil (approved):

.github/workflows/docker_push.yml
leonardBang: +1 to keep 2.0.x as it is a LTS version and discuss this topic in a new thread

leonardBang (approved):

PR #273, opened 2026-07-09
STRY64301574: Update log4j to fix CVE-2025-68161

sean-campbell-now:


snuyanzin (changes requested):
Merge commit is not allowed
flink-docker is not the place where dependencies like log4j should be updated

